PR #153, opened 2015-03-13
silence gcc's missing-field-initializers warnings

darealshinji:


wm4:
This should probably use ``{}`` instead (C++11 I think).

dubhater:
It's a bug in gcc 4.9. Hopefully they will fix it eventually and ``{ 0 }`` will work without warnings again.

myrsloik:
{0} changed to {} everywhere in the c++ code

PR #171, opened 2015-06-23
Generic: correct rounding for Deflate/Inflate.

IFeelBloated:
correct rounding for int type


dubhater:
Please use descriptive commit messages. For example, "Generic: correct rounding for Deflate/Inflate".

From the [git book](https://git-scm.com/book/ch5-2.html#Commit-Guidelines):

> As a general rule, your messages should start with a single line that’s no more than about 50 characters and that describes the changeset concisely, followed by a blank line, followed by a more detailed explanation.

(There is no need for a detailed explanation this time.)

IFeelBloated:
okay, noted :)

PR #179, opened 2015-09-24
Assvapour Subtitle Improvements

splinter98:
Two features/fixes in assvapour.Subtitle

1) Enable use of animation tags for Subtitle (such as {\fad} or {\K})
2) Enable adjusting the start and end time on the Subtitle

### DIFF
--- a/src/filters/assvapour/assvapour.c
+++ b/src/filters/assvapour/assvapour.c
@@ -19,6 +19,7 @@
 #include <string.h>
 #include <ass/ass.h>
 #include <time.h>
+#include <inttypes.h>
 
 #include "VapourSynth.h"
 #include "VSHelper.h"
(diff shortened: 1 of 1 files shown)


splinter98:
That should protect against overflowing in a very rare case someone uses a crazy slow frame rate.

myrsloik:
Would you mind updating the documentation too?

splinter98:
No problem I will get on that.

dubhater:
Shouldn't *end* default to numFrames - 1, for correct animations and whatnot?

splinter98:
Because it gets converted to time and not framed, if it was numFrames-1 the last frame wouldn't have subtitle (though it may due to rounding errors)

myrsloik:
I think splinter98 is right

PR #199, opened 2015-12-19
Remove PlaneDifference register function

4re:


myrsloik:
I simply forgot to sync my other changes yesterday evening, should compile now
beware of piles of untested code

PR #218, opened 2016-03-12
vsresize: add support for field-based and interlaced video

### LINKED ISSUE #216: Resizers don't like interlaced video
Even operations like bit depth conversion fail with an error if the input has the _FieldBased property and it's not 0.

sekrit-twc:
Fixes #216

### DIFF
--- a/src/core/vsresize.cpp
+++ b/src/core/vsresize.cpp
@@ -18,6 +18,7 @@
 * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
 */
 
+#include <algorithm>
 #include <limits>
 #include <memory>
 #include <stdexcept>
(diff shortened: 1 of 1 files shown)


src/core/vsresize.cpp
```diff
@@ -503,10 +533,24 @@ namespace {
...
+        void init_pack(const vszimgxx::FilterGraph &graph, VSFrameRef *frame, const zimg_image_format &format, bool interlaced,
+                       const VSFormat *vsformat, VSCore *core, const VSAPI *vsapi) {
             init_common(graph, true, format, vsformat, core, vsapi);
             import_frame_as_buffer(frame, &m_plane_buf.m, ZIMG_BUFFER_MAX, vsapi);
 
+            if (interlaced) {
+                unsigned order = format.field_parity == ZIMG_FIELD_BOTTOM ? 1 : 0;
+
+                // CompatBGR32 is upside-down.
+                if (vsformat->colorFamily == cmCompat)
```
dubhater: This will assume YUY2 is upside down too, right?

sekrit-twc: It's a bug

PR #235, opened 2016-08-22
Add Hysteresis filter

HolyWu:


src/core/genericfilters.cpp
```diff
@@ -2645,67 +2651,262 @@ static void VS_CC levelsCreate(const VSMap *in, VSMap *out, void *userData, VSCo
...
+    } else if (activationReason == arAllFramesReady) {
+        const VSFrameRef * src1 = vsapi->getFrameFilter(n, d->node1, frameCtx);
+        const VSFrameRef * src2 = vsapi->getFrameFilter(n, d->node2, frameCtx);
+        const VSFrameRef * fr[] { d->process[0] ? nullptr : src1, d->process[1] ? nullptr : src1, d->process[2] ? nullptr : src1 };
+        const int pl[] { 0, 1, 2 };
+        VSFrameRef * dst = vsapi->newVideoFrame2(d->vi->format, d->vi->width, d->vi->height, fr, pl, src1, core);
+
+        auto threadId = std::this_thread::get_id();
+        if (!d->label.count(threadId)) {
+            uint8_t * label = vs_aligned_malloc<uint8_t>(d->vi->width * d->vi->height, 32);
```
dubhater: What's the point of using aligned malloc if only the first line is aligned?

src/core/genericfilters.cpp
```diff
@@ -2645,67 +2651,262 @@ static void VS_CC levelsCreate(const VSMap *in, VSMap *out, void *userData, VSCo
...
+    VSNodeRef * node1, * node2;
+    const VSVideoInfo * vi;
+    bool process[3];
+    uint16_t peak;
+    float lower[3], upper[3];
+    std::unordered_map<std::thread::id, uint8_t *> label;
+};
+
+template<typename T>
+static void process_plane_hysteresis(const VSFrameRef * src1, const VSFrameRef * src2, VSFrameRef * dst, const HysteresisData * d, const VSAPI * vsapi) VS_NOEXCEPT {
```
dubhater: You mean "process_frame_hysteresis"?

doc/functions/hysteresis.rst
```diff
@@ -0,0 +1,15 @@
+Hysteresis
```
myrsloik: Just a note on adding it to the core. I think it's a bit too specific to do that but I'm currently working on a plugin with useful extra functions I'd like to include it in. Where I'd include small useful things nobody can be bothered with elsewhere but are still good or replaces some avisynth functionality. I'm currently working on a function to average frames to finally evict temporalsoften from all scripts as well.

src/core/genericfilters.cpp
HolyWu: Ha. I really got used to use vs_aligned_malloc and wasn't aware of that.

myrsloik:
Released in my miscfilter pile of stuff

PR #265, opened 2017-01-12
Introspection Pt. 3

stuxcrystal:
Contains the following changes:

1. Shows the name of the plugin in a way that the help function will *hopefully* show it. (If not IPython does.) 
2. Show the available namespaces in vapoursynth.Core and vapoursynth.VideoNode
3. Shows the correct function names if no environment_id was found.

Off-Topic:
I was wondering if splitting the vapoursynth.pyx files would make the code more readable? What is your opinion @myrsloik


myrsloik:
I'd rather not split it, it'll just cause more troubles with cython later. Other improvements like this one are interesting though.

PR #266, opened 2017-01-13
Actually add the 'name' description of the plugin

Kamekameha:
Previously it just took the description of the first plugin found.

Unfortunately, I couldn't find anything similar to a "namespace to identifier"/"get identifier from plugin"/"whatever" function in the public API (read: too lazy to look for thoroughly), so I did an extra loop through the available plugins. Otherwise, a simple `core.get_plugins()[identifier_string_or_something]['name']` would have been enough.


myrsloik:
I don't think there's an easy way to get that from the API. The API assumes that you only ever want to go from namespace/id to a plugin, not the other way around.

PR #271, opened 2017-01-17
Let VideoProps behave like a dict

stuxcrystal:
vapoursynth.VideoProps now behaves like a dictionary while keeping backwards compatibility.

There are two changes worth documenting:

1. The default-parameter of VideoProps.setdefault will be zero by default instead of None as the FrameProp-Structure does not support None-Types
2. VideoProps.copy will return a PyDict-Object instead a copy of itself.

Let's paste examples here (pasted from multiple IPython sessions):
Also serves as a simple test :P

```python
In [1]: import vapoursynth; c=vapoursynth.get_core(); bc = c.std.BlankClip(); f=bc.get_frame(0); f2=f.copy()

In [3]: dict(f2.props)
Out[3]: {'_DurationDen': 24, '_DurationNum': 1}

In [4]: f2.props._DurationDen
Out[4]: 24

In [5]: f2.props['_DurationNum']
Out[5]: 1

In [6]: f2.props['_DurationNum'] = 2

In [6]: del f2.props['_DurationDen']

In [7]: f2.props
Out[7]: <vapoursynth.VideoProps {'_DurationNum': 1}>

In [7]: dict(f2.props)
Out[7]: {'_DurationDen': 24, '_DurationNum': 2}

In [8]: f2.props.get("_Matrix")

In [9]: f2.props.get("_Matrix", "709")
Out[9]: '709'

In [10]: f2.props.keys()
Out[10]: <generator at 0x12d11c06598>

In [11]: tuple(f2.props.keys())
Out[11]: ('_DurationDen', '_DurationNum')

In [12]: tuple(f2.props.values())
Out[12]: (24, 2)

In [13]: tuple(f2.props.items())
Out[13]: (('_DurationDen', 24), ('_DurationNum', 2))

In [14]: tuple(f2.props)
Out[14]: ('_DurationDen', '_DurationNum')

In [15]: del f2.props._DurationDen

In [16]: f2.props._DurationDen = 24

In [20]: f2.props
Out[20]: <vapoursynth.VideoProps {'_DurationDen': 24, '_DurationNum': 2}>

In [22]: f2.props.setdefault('_DurationDen', 12)
Out[22]: 24

In [23]: f2.props.setdefault('_DurationDen2', 12)
Out[23]: 12

In [25]: f2.props
Out[25]: <vapoursynth.VideoProps {'_DurationDen': 24, '_DurationDen2': 12, '_DurationNum': 2}>

In [27]: f2.props.clear()

In [29]: f2.props
Out[29]: <vapoursynth.VideoProps {}>

In [2]: f2.props.clear()

In [3]: f2.props.update(f.props)

In [5]: f2.props
Out[5]: <vapoursynth.VideoProps {'_DurationDen': 24, '_DurationNum': 1}>

In [10]: len(f2.props)
Out[10]: 2
```

### DIFF
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -784,7 +784,7 @@ cdef class VideoProps(object):
         if 0 < len(args) < 2:
             args = args[0]
             if not isinstance(args, dict):
-                args = dict(args[0])
+                args = dict(args)
         elif len(args) > 1:
             raise TypeError("update takes 1 positional argument but %d was given" % len(args))
         else:
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -23,12 +23,14 @@ from cython cimport view
 from libc.stdint cimport intptr_t, uint16_t, uint32_t
 from cpython.ref cimport Py_INCREF, Py_DECREF
 import os
+import abc
 import ctypes
 import threading
 import traceback
 import gc
 import sys
 import inspect
+from collections.abc import Mapping
 
 # Ensure that the import doesn't fail
 # if typing is not available on the python installation.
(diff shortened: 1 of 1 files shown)


myrsloik:
So both the old and the new way works then so no existing scripts break?
The only thing I can spot is the copy method, that one could bring a lot of pain when not used properly. But I guess users have to beware of that.

stuxcrystal:
I thought that too, that's why I explicitly stated that behaviour in the PR.

I decided on that because the VideoProp-Object references the actual frame and since the copy-method is defined as a shallow-copy users will expect to be able to modify the copy. Thus I decided on returning an actual dictionary which the user can then modify.

PR #273, opened 2017-01-17
 Implement PEP 435's enums where applicable

Kamekameha:
Cython has been supporting these for cpdef'd enums for quite a while (since 0.24 actually), so why not?

The resulting enums are `int` instances so it "shouldn't" break any user code.

### DIFF
--- a/src/cython/vapoursynth.pxd
+++ b/src/cython/vapoursynth.pxd
@@ -38,64 +38,64 @@ cdef extern from "include/VapourSynth.h" nogil:
     ctypedef struct VSFrameContext:
         pass
 
-    cdef enum VSColorFamily:
-        cmGray  = 1000000
-        cmRGB   = 2000000
-        cmYUV   = 3000000
-        cmYCoCg = 4000000
-        cmCompat= 9000000
+    cpdef enum ColorFamily "VSColorFamily":
+        GRAY "cmGray"
+        RGB "cmRGB"
+        YUV "cmYUV"
+        YCOCG "cmYCoCg"
+        COMPAT "cmCompat"
 
-    cdef enum VSSampleType:
-        stInteger = 0
-        stFloat   = 1
+    cpdef enum SampleType "VSSampleType":
+        INTEGER "stInteger"
+        FLOAT "stFloat"
 
-    cdef enum VSPresetFormat:
-        pfNone = 0
+    cpdef enum PresetFormat "VSPresetFormat":
+        NONE "pfNone"
 
-        pfGray8 = cmGray + 10
-        pfGray16
+        GRAY8 "pfGray8"
+        GRAY16 "pfGray16"
 
-        pfGrayH
-        pfGrayS
+        GRAYH "pfGrayH"
+        GRAYS "pfGrayS"
 
-        pfYUV420P8 = cmYUV + 10
-        pfYUV422P8
-        pfYUV444P8
-        pfYUV410P8
-        pfYUV411P8
-        pfYUV440P8
+        YUV420P8 "pfYUV420P8"
+        YUV422P8 "pfYUV422P8"
+        YUV444P8 "pfYUV444P8"
+        YUV410P8 "pfYUV410P8"
+        YUV411P8 "pfYUV411P8"
+        YUV440P8 "pfYUV440P8"
 
-        pfYUV420P9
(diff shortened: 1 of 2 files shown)


myrsloik:
a7ea620 needs to have long added to the isinstace tuple, unless you can prove the code doesn't compile properly for python 2.7 already
But bbdb022 I'll reject since it needlessly requires a much, much newer cython for no real benefit

Kamekameha:
Ah, I purposely ignored the long type since the docs here explicitly says that it requires Python 3, other than that yeah I should have left it.
Also, Cython doesn't write special code to the generated C file for declarations within a `cpef extern from` block, so specifying the values of constants isn't really required, the value should already exist in the included header. As for the cpdef enum thingy you could take a look at [Cython/Utility/CpdefEnums.pyx](https://github.com/cython/cython/blob/d02cc4c5d831da27cd871cbb3feaf8bea72ec0c0/Cython/Utility/CpdefEnums.pyx) to see how it works. All in all, this compiled fine for me and the values are definitely there so I'm curious about which errors you're getting.
And finally, isn't `pip` a viable way to update Cython? I don't see any issue with requiring a newer version considering that Cython itself is not needed for the compiled module at the end of the day.

Kamekameha:
Oh, and almost forgot, the recently added signature stuff, more specifically `Parameter` and `Signature` classes of the `inspect` since the `typing` module is already worked around, aren't really Python 2 compatible, so support for that should be a different concern. In general, Python 2 support needs some work.

stuxcrystal:
@Kamekameha Actually, the Signature-stuff shouldn't cause an issue on older Python-versions as they are encapsulated in the construct_parameter and construct_signature functions. It doesn't get called on systems where the `typing` is not available (this includes Python 2). The inspect-module itself is available since time immemorial (aka. Python 2.1).

Kamekameha:
Yes, that's why I said:

> more specifically the `Parameter` and `Signature` classes of the `inspect` module

which aren't available in Python 2, while there's a `typing` module backported on PyPI by the very Python Devs that implemented PEP 484, so that existence check could easily be bypassed by simply installing that backport, and that might happen without users even knowing they have that installed.
Idk, I'm just saying. As it is, the python module of vapoursynth doesn't really works on Python 2 without some (rather minor) changes.

myrsloik:
I've now verified that it's horribly broken on 2.x python and there's no hope for it. So I guess we can go ahead with the cpdef stuff.

Kamekameha:
Broken how badly? Because I remember the changes I did locally quite a while ago to make it work weren't really that much. I mean, Cython does a pretty good job already in making the resulting C code Python 2/3 compatible in most cases.

myrsloik:
After poking modules that don't exist it complains about str/unicode object type mismatch. I mean 2.x support is definitely doable but I'm now going to officially kill the idea.

Kamekameha:
Jesus, Idek what I'm doing while rebasing

myrsloik:
https://github.com/vapoursynth/vapoursynth/pull/273/commits/ff94b3424567ff84b64917aafca03b4258d73daa#diff-38001efdc649c129c2406c34450f34d3R147
^how does that figure out that arError is -1? The other ones I can kinda guess how it works but that one I'm suspicious of

Kamekameha:
Because like I said, Cython doesn't generate extra code for `cdef extern from "yourheader.h"` blocks. Look at [external_C_code.html](https://cython.readthedocs.io/en/latest/src/userguide/external_C_code.html#referencing-c-header-files). It simply adds `#include "yourheader.h"` in the generated C code, Cython doesn't need to know what value constants have; any reference, in this case for example, of arError should have already been defined on VapourSynth.h (which it is).

Kamekameha:
Regarding the str/unicode mismatch, it might help adding `# cython: language_level=3` at the top of the script and work from there. I could try and make some commits if there's still interest in supporting Python 2, should be simple honestly. Most of the compatibility work is already done by Cython.

myrsloik:
No, python2 isn't relevant anymore. It kinda was back when I started with the experimental Python 3.1 but now everything of value is ported. At least everything you could reasonably want to combine with video stuff

Kamekameha:
Alright. Anyway, let me know if there's anything this PR needs to change.

PR #277, opened 2017-01-18
Add frames-function to clip

stuxcrystal:
For easier access to frames should you want to access successive frames.


dubhater:
How is this different from calling get_frame in a loop?

myrsloik:
It's kinda the same but all python stuff has iterators

PR #279, opened 2017-01-20
add get_frames_async for asynchronous frame retrieval

stuxcrystal:
Renamed the original `CallbackData`-class to CallbackDataOutput and done similar things to the callback function.

### DIFF
--- a/doc/pythonreference.rst
+++ b/doc/pythonreference.rst
@@ -287,6 +287,42 @@ Classes and Functions
 
       Returns a VideoFrame from position n.
 
+   .. py:method:: get_frame_async(n)
+
+      Returns a concurrent.futures.Future-object which result will be a VideoFrame instance or sets the
+      exception thrown when rendering the frame.
+
+      *The future will always be in the running or completed state*
+
+      :param n: The frame number
+
+   .. py:method:: get_frame_async_raw(n, cb: callable)
+
+      First form of this method. It will call the callback from another thread as soon as the frame is rendered.
+
+      The `result`-value passed to the callback will either be a VideoFrame-instance on success or a Error-instance
+      on failure.
+
+      *This method is intended for glue code. For normal use, use get_frame_async instead.*
+
+      :param n: The frame number
+      :param cb: A callback in the form `cb(node, n, result)`
+
+   .. py:method:: get_frame_async_raw(n, cb: Future[, wrapper: callable = None])
+
+      Second form of this method. It will take a Future-like object (including asyncio.Future or similar)
+      and set its result or exception according to the result of the function.
+
+      The optional `wrapper`-parameter is intended for calls like asyncio.EventLoop.call_soon_threadsafe in which
+      all calls to its future-object must be wrapped.
+
+      *This method is intended for glue code. For normal use, use get_frame_async instead.*
+
+      :param n: The frame number
+      :param cb: The future-object whose result will be set.
+      :param wrapper: A wrapper-callback which is responsible for moving the result across thread boundaries. If not
+                      given, the result of the future will be set in a random thread.
+
    .. py:method:: set_output(index = 0)
 
       Set the clip to be accessible for output. This is the standard way to
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -23,13 +23,13 @@ from cython cimport view
 from libc.stdint cimport intptr_t, uint16_t, uint32_t
 from cpython.ref cimport Py_INCREF, Py_DECREF
 import os
-import abc
 import ctypes
 import threading
 import traceback
 import gc
 import sys
 import inspect
+from concurrent.futures import Future
 from collections.abc import Mapping
 
 # Ensure that the import doesn't fail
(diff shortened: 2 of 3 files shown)


dubhater:
How do I do this with get_frame_async?
```python
src = core.tdm.IsCombed(src, cthresh=8)

with open("combed.txt", "w") as f:
    for i in range(src.num_frames):
        if src.get_frame(i).props["_Combed"] == 1:
            f.write("{}\n".format(i))
```

Kamekameha:
I haven't tested this myself but probably like this,
```python
from concurrent.futures import wait

src = core.tdm.IsCombed(src, cthresh=8)

with open('combed.txt', 'w') as f:
    futures = []

    for x in range(len(src)):

        def write_on_combed(future, number=x):
            frame = future.result()
            if frame.props['_Combed']:
                print(number, file=f)

        fut = src.get_frame_async(x)
        fut.add_done_callback(write_on_combed)
        futures.append(fut)

    done, not_done = wait(futures)
```
Although I got to say, and I meant to also comment about it on #281, that I couldn't find a real need for this `get_frame_async` method. You see, pretty much the same could've been achieved through already existing concurrent libraries from the standard library. The thing is, `get_frame` releases the GIL where needed and so it lets other Python threads to execute in the meantime, therefore not blocking Python itself.

Using the very same `concurrent.futures` module wouldn't even look that much different from the snippet above (I haven't tested this either):
```python
from concurrent.futures import ThreadPoolExecutor

src = core.tdm.IsCombed(src, cthresh=8)

with open('combed.txt', 'w') as f, ThreadPoolExecutor() as executor:
    for x in range(len(src)):

        def write_on_combed(future, number=x):
            frame = future.result()
            if frame.props['_Combed']:
                print(number, file=f)

        future = executor.submit(src.get_frame, x)
        future.add_done_callback(write_on_combed)
```
Additionally, according to the documentation, direct construction of `Future` objects and usage of some of it methods aren't exactly public API and are meant to be used by no other than the executor itself.

In short, there are plenty of ways to get frames without blocking Python that didn't need any modification to the vapoursynth code base.

stuxcrystal:
`VideoNode.get_frame_async` is a rather low-level primitive to request frames. It does not save you from creating 20k+ requests for frames filling up your ram.

That being said, the main problem using `get_frame_async`, is its memory-use when not using a semaphore or similar techniques as this would just dump all requests into the queue (or whatever vapoursynth uses internally) and in turn essentially loading frames into ram.

https://github.com/Infiziert90/getnative/blob/master/getnative.py#L143

while this code AsyncIO, all primitives used in that particular code have equivalents (with the same name) in concurrent.futures.

I propose this solution for using `get_frame_async` (I didn't get around to testing it yet):
```py
import vapoursynth as vs
from concurrent.futures import wait, FIRST_COMPLETED, ALL_COMPLETED

def each_frame(clip: vs.VideoNode, prefetch=None, *, core: vs.Core = None):
    if core is None:
       core = vs.core.core
    if prefetch is None:
       prefetch = core.num_thread

    frames = {}
    current = 0
    futures = {}

    def wait_for_finished_frame():
       if futures:
           # Wait for a frame to finish.
           done, not_done = wait(futures.values(), return_when=FIRST_COMPLETED)
           finished_frameno = futures.pop(done[0])
           frames[finished_frameno] = done[0]

       # Yield the current frame
       if current in frames:
           cf = current
           current+=1
           return frames.pop(cf)


    for i in range(len(clip)):
       # Actually request a new frame.
       future = clip.get_frame_async(i)
       futures[future] = i
       if len(futures) < prefetch:
           continue
       result = wait_for_finished_frame()
       if result is not None:
           yield result
       
    while futures and frames:
       yield wait_for_finished_frame()

with open("combed.txt", "w") as f:
    f.write("\n".join(
        str(i)
        for i, f in enumerate(
            fut.result()
            for fut in
            each_frame(src)
        )
        if f.props['_Combed']==1
     ))
```

stuxcrystal:
@Kamekameha I actually created this function because I really don't like creating new threads (or using it's thread-pool-executor) in AsyncIO.

I feel that it is a waste of resources creating threads solely for waiting on other threads.

dubhater:
This is probably not actually good, but what if you do something like this?
```python
for i in range(src.num_frames):
    if i % core.num_threads == 0:
        for x in range(i, min(i + core.num_threads, src.num_frames)):
            src.get_frame_async(x)

    frame = src.get_frame(i)
    # do stuff with frame
```

Kamekameha:
@stuxcrystal I thought so, but does that really matters at the end of the day?

If anything it complicates things a bit;
* Doesn't really simplify API use-cases;
* Requires somewhat manual tracking of `Future` instances;
* Raises issues like #281 – leaving the method (after moving the `concurrent.futures` import around) in a "use at your own risk" state when combined with vspipe;
* Adds some unnecessary maintenance responsibility for a non-idempotent method;
* Risks calling added callbacks in non-Python threads whenever that matters (it usually doesn't);
* And, if interest ever pops out, breaks support for execution in sub-interpreters.

Being the last two not exactly specific to this PR.

Whereas simply letting the user use their favorite concurrent library to get frames asynchronously, with a note that vspipe doesn't do well together with the `signal` module, avoids pretty much all of these issues and shifts any related responsibility towards them. That without needing to make a single change to vapoursynth.

And I'm aware that combining `get_frame_async` or anything similar together with vspipe isn't a likely to happen scenario, but it's still quite an easy one to trigger.

Of course, I may as well be exaggerating given that nobody other than me now has complained so far, and that I'm not particularly interested in what vspipe supports or not. Just pointing things out.

PR #282, opened 2017-02-05
Add proxy object for the Core class.

stuxcrystal:
At the moment, most VapourSynth Libraries are currently required to do essentially this:

```python
import vapoursynth

def useful_function(arg1=240):
    core = vapoursynth.get_core()
    return core.std.BlankClip(length=arg1)
```

This change reduces the redundancy of calling `get_core` for every function by adding a simple proxy-object which will resolve the current Core-instance on attribute access.

The code above would change into

```python
from vapoursynth import core

def useful_function(arg1=240):
    return core.std.BlankClip(length=arg1)
```

### DIFF
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -1388,6 +1388,7 @@ cdef class Core(object):
         s += '\tAdd Cache: ' + str(self.add_cache) + '\n'
         s += '\tAccept Lowercase: ' + str(self.accept_lowercase) + '\n'
         return s
+        
 
 cdef Core createCore():
     cdef Core instance = Core.__new__(Core)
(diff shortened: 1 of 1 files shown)


dubhater:
No documentation?

myrsloik:
The point is that no documentation is needed, it just abstracts away the core reference quirk. Or I guess removing that quirk from the docs should be done...

stuxcrystal:
Actually it should be documented. Maybe as the new standard way to get a core instance.

The documentation of the `get_core()`-function should now include a remark that it allows for more control over how a new core instance should be configured.

myrsloik:
Actually setting the number of threads in get_core is obsolete too, I already made everything configurable as core attributes long ago. After this change get_core() serves no real purpose.

PR #285, opened 2017-02-14
Manually implement the buffer protocol for VideoFrame's planes

Kamekameha:
This is an experimental attempt to fix the unconditionally writable memviews that `get_{read|write}_array` returns. Unfortunately, it obviously doesn't have all the functionalities that cython's memoryviews gives but to my relief all vapoursynth snippets I've seen that used `get_{read|write}_array` were also accompanied with numpy, so it doesn't seem to me that a lot of users were using these functionalities that often.

### DIFF
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -352,78 +352,52 @@ cdef void __stdcall frameDoneCallbackRaw(void *data, const VSFrameRef *f, int n,
             Py_DECREF(d)
 
 
-cdef void __stdcall frameDoneCallbackOutput(void *data, const VSFrameRef *f, int n, VSNodeRef *node, const char *errormsg) nogil:
-    cdef int pitch
-    cdef const uint8_t *readptr
-    cdef const VSFormat *fi
-    cdef int row_size
-    cdef int height
-    cdef char err[512]
-    cdef int p
-    cdef int y
+cdef void __stdcall frameDoneCallbackOutput(void *data, const VSFrameRef *f, int n, VSNodeRef *node, const char *errormsg) with gil:
+    cdef VideoFrame frame_obj
+    cdef VideoPlane plane
 
-    with gil:
-        d = <CallbackData>data
-        d.completed = d.completed + 1
-        
-        if f == NULL:
-            d.total = d.requested
-            if errormsg == NULL:
-                d.error = 'Failed to retrieve frame ' + str(n)
-            else:
-                d.error = 'Failed to retrieve frame ' + str(n) + ' with error: ' + errormsg.decode('utf-8')
-            d.output = d.output + 1
+    cdef CallbackData d = <CallbackData>data
+    d.completed += 1
 
+    if f == NULL:
+        d.total = d.requested
+        if errormsg == NULL:
+            d.error = 'Failed to retrieve frame ' + str(n)
         else:
-            d.reorder[n] = createFramePtr(f, d.funcs)
+            d.error = 'Failed to retrieve frame ' + str(n) + ' with error: ' + errormsg.decode('utf-8')
+        d.output += 1
+    else:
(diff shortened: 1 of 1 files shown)


myrsloik:
I guess this is ok but I have to stare at it a bit more, I'm no expert at this kind of python stuff...

myrsloik:
You don't seem to handle half precision at all, it'll be incorrectly treated as single precision?

splinter98:
To be fair neither does my original code. FP16 is a bit of a pain to implement due to the lack of type to cast it against.

We probably should write a VS_HALFFLOAT type that can represent and manipulate this type, for devices that don't have hardware/compiler types with fast paths for cpus/devices that support them. Then we can cast half floats here by importing the type into cython. This would also benefit plugin developers to enable them to treat half float in a similar way they do to single precision. (This may be there in VS already but I couldn't find it not as a custom type at least)

myrsloik:
Or just expose it has int16 or whatever. That's the type that gets abused internally for it anyway in a few places. Either way it should be handled somehow... errors are better than corrupt data

Kamekameha:
Oh, yeah well, I just copied the same checks that were done in get_read_array/get_write_array. I guess I could raise an exception in such cases. Any ideas?

Kamekameha:
Never mind. Done.

myrsloik:
You call Py_INCREF in one place but there's no decref? is this right?

Kamekameha:
Yes, `PyTuple_SetItem` (and its macro form) steals a reference from, in this case, the plane object.

Kamekameha:
I just noticed this PR got reverted, may I know why? And sorry beforehand if I missed or fucked up something I should have noticed.

myrsloik:
It caused memory leaks. Even a simple thing like:
import vapoursynth as vs
 
c = vs.get_core(add_cache=False)
 
src = c.std.BlankClip(length=500, width=1920, height=1080)
 
for i in range(src.num_frames):
    src.get_frame(i)

Would cause problems. Reverting the commits fixed it

Kamekameha:
Hm, weird. It seems more like they aren't garbage collected when appropriate, maybe there's a reference cycle between the `planes` tuple and the `VideoFrame` instance that the gc can't break? Anyway, I don't think reverting the whole thing was necessary, just removing the `planes` attribute would probably have been good enough, `VideoPlane` objects could be instantiated separately or whatever.

myrsloik:
That's the thing. It takes me even more time to systematically debug it, especially since I didn't write it or studied the relevant APIs. So for now I reverted it until I have time to do something about it. A debugged version is of course welcome.

Kamekameha:
Ah, yeah, I completely understand that. Anyway, it was definitely that reference cycle between these two that caused those memleaks, so it's pretty much just removing that `planes` attribute from `VideoFrame` (or at least I hope so :^) ), do you want me to do the relevant changes or do you prefer to look further into this before anything?

myrsloik:
You get to do it. Just make sure to test it for memory leaks

Kamekameha:
Well, the only "testing" I did was running the gc in debug mode and check for objects that became unreachable (but still free-able) after running the snippet you posted above, and there were indeed unreachable `VideoFrame` objects with the tuple attribute while there weren't when I simply moved the attribute logic out of the initialization step (and of course changing the snippet to request all planes on each frame iteration, just in case). Would that be good enough testing?

Kamekameha:
Never mind, made a PR either way. Also,

> That's the thing. It takes me even more time to systematically debug it, especially since I didn't write it or studied the relevant APIs. So for now I reverted it until I have time to do something about it. A debugged version is of course welcome.

While I say I completely understand that, we still could've had this very same convo without reverting the whole PR, reverting is not exactly the same as fixing. As we can see I tried to address the issue as soon as I noticed the changes, in case that expecting a prompt answer here was the concern. Otherwise we're going to see unnecessary commits here and there.

PR #287, opened 2017-02-15
Use a Fraction instance for the FPS

Kamekameha:
The idea into this is to deprecate/remove the `fps_num`/`fps_den` attributes of `VideoNode` and use this instead. If anyone needs the FPS values at the C level or something they could simply get them from the `vi` member.

Btw, you're free to ignore the 2nd commit if it's not that representative as a string for the `__str__` method.

### DIFF
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -1217,12 +1217,7 @@ cdef class VideoNode(object):
 
         s += '\tNum Frames: ' + str(self.num_frames) + '\n'
 
-        if not self.fps_num or not self.fps_den:
-            s += '\tFPS Num: dynamic\n'
-            s += '\tFPS Den: dynamic\n'
-        else:
-            s += '\tFPS Num: ' + str(self.fps_num) + '\n'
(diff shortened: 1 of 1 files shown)


myrsloik:
Needs to be documented

dubhater:
What's the point of this change?

Kamekameha:
Not like that should really matter given that the check in [src/cython/vapoursynth.pyx#L1217](https://github.com/vapoursynth/vapoursynth/blob/bf9677923d6af725bd8afdbe9f0b69f683264f8b/src/cython/vapoursynth.pyx#L1217) suggested that it's dynamic if either the numerator or denominator is 0. I mean, I could change it to None if really necessary, but a fraction of 0/1 already suggests it's an invalid framerate, at least in this case (and I actually prefer a dummy fraction than a None, it's more consistent, but that's just me).

myrsloik:
I assume the idea is to use the intended python class for rational numbers. Which I probably should've done from the start...

Kamekameha:
Pretty much.

Issue #300, opened 2017-03-30
get_read_ptr outputs corrupt data after shape manipulation


invisiblearts:
Edit: using python highlight in the code snippet.

I was trying to extract the frame data to numpy arrays. `get_read_array` works fine but I would like to make it faster. `get_read_ptr` is (about a hundred times) faster for reading, based on a simple test, but I met the following issue (please see script).

```python
import vapoursynth as vs
import numpy as np
import ctypes
import mvsfunc

# CHW if not HWC
def get_array_ctype(src, frame_num, HWC=True):
    assert(isinstance(src, vs.VideoNode))
    assert(isinstance(frame_num, int))
    assert(isinstance(HWC, bool))
    

    frame = src.get_frame(frame_num)
    frame = frame.copy() # explicitly copying to avoid potential memory management conflicts
    num_planes = frame.format.num_planes

    out_list = []
    
    st = frame.format.sample_type
    bps = frame.format.bytes_per_sample

    if st == vs.INTEGER:
        if bps == 1:
            dtype = ctypes.c_uint8
        elif bps == 2:
            dtype = ctypes.c_uint16
        else:
            raise ValueError('Wrong sample type!')
    elif st == vs.FLOAT:
        if bps == 2:
            raise ValueError('Half sample is not natively supported! Please resample it.')
        elif bps == 4:
            dtype = ctypes.c_float
        else:
            raise ValueError('Wrong sample type!')


    for i in range(num_planes):
        arr = np.ctypeslib.as_array((dtype * frame.width * frame.height).from_address(frame.get_read_ptr(i).value))
        out_list.append(arr.copy()) # still, (reduntantly) explicitly copying to avoid potential memory management conflicts

    # check to prevent subsampling issues
    shape = arr.shape
    for arr in out_list:
        assert(arr.shape == shape)

    if HWC:
        return np.dstack(out_list)
    else:
        return np.array(out_list)

def get_array(src, frame_num, HWC=True):
    assert(isinstance(src, vs.VideoNode))
    assert(isinstance(frame_num, int))
    assert(isinstance(HWC, bool))
    

    frame = src.get_frame(frame_num)
    num_planes = frame.format.num_planes

    out_list = []
    
    for i in range(num_planes):
        arr = np.array(frame.get_read_array(i))
        out_list.append(arr)

    # check to prevent subsampling issues
    shape = arr.shape
    for arr in out_list:
        assert(arr.shape == shape)

    if HWC:
[...]

PR #303, opened 2017-03-31
More plane buffer

Kamekameha:
This addresses a memory leak noted in https://github.com/vapoursynth/vapoursynth/pull/285#issuecomment-290371636 due to a reference cycle between a `VideoFrame` instance and its `planes` tuple attribute that kept the object alive through the whole Python execution (which I should've seen that coming tbh).

### DIFF
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -352,78 +352,52 @@ cdef void __stdcall frameDoneCallbackRaw(void *data, const VSFrameRef *f, int n,
             Py_DECREF(d)
 
 
-cdef void __stdcall frameDoneCallbackOutput(void *data, const VSFrameRef *f, int n, VSNodeRef *node, const char *errormsg) nogil:
-    cdef int pitch
-    cdef const uint8_t *readptr
-    cdef const VSFormat *fi
-    cdef int row_size
-    cdef int height
-    cdef char err[512]
-    cdef int p
-    cdef int y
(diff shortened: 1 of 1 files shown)


Kamekameha:
Oh, and BTW, this `VideoPlane` stuff should (or hopefully, sometime in the near future :) ) deprecate the current `get_{read,write}_{ptr,array}` stuff for multiple reasons (well, IMO of course):
- For `get_{read,write}_ptr`, there couldn't possibly be any faster approach to correctly retrieve the underlying buffer in Python space than through `VideoPlane` (mostly stating this because of code like in https://github.com/vapoursynth/vapoursynth/issues/300#issue-218108195, so FTR: https://gist.github.com/Kamekameha/68f7b6291b1290147e0241b0cd350f01);
- For `get_{read,write}_array`, typed memoryviews are actually more useful in Cython space since they do not require the GIL after created, so one could use those in interesting places while still having a numpy-like interface, in Python space one can't really take advantage of this;
- In all cases the resulting array is writable no matter which of these functions is called;
- And in all cases as well it's pretty much repeated code with some checks added.

I'm all open (and would love) to suggestions and opinions on this.

stuxcrystal:
https://github.com/stuxcrystal/yuuno/blob/development/yuuno/core/converter/vs_conv.py#L36

This is the fastest approach I could find. But yes, your approach would be faster now.

myrsloik:
I thought about it and realized that I don't have any particular strong opinions on these python details.

Issue #283, opened 2017-02-06
Make vapoursynth a standard Python package


chengyuhui:
The current installation method involves running an installer or copying files into python directory, which makes it difficult to deploy applications and upgrading python automatically.

Is it possible to use setuptools to do the packaging?

stuxcrystal:
The main problem is here: `setup.py` L12-17
```python
library_dirs = [curdir, "build"]
if is_win:
    if is_64:
        library_dirs.append(join("msvc_project", "x64", "Release"))
    else:
        library_dirs.append(join("msvc_project", "Release"))
```

It hardcodes the places where the vapoursynth.lib is being searched.

Kamekameha:
Yeah, agree. IMO the `setup.py` script shouldn't have any hardcoded paths and let the user building the python module specify the include and library directories manually (with the `-I`/`-L`  options accordingly, or in `CFLAGS`/`LDFLAGS` if the system supports it). Though that basically means one extra thing to keep in mind for those distributing this, but that's essentially how all CPython extensions are distributed on PyPI.

stuxcrystal:
I chatted with Myrsloik yesterday. I am going to add a bit of code adding the discovery of the vapoursynth-sdk on windows. On other systems, I will just assume it will be on the path.

chengyuhui:
Actually, I'm thinking of moving everything (vapoursynth itself, plugins, scripts, etc.) onto PyPI.

AFAIK, not many vapoursynth users are real Python developers, but we can still benefit from the Python ecosystem. 

e.g.: Once this is done, it'll be easier to implement #52 

Another problem: is there any real difference between plugins and core plugins? Maybe we can distribute them separately.

myrsloik:
Core plugins are always guaranteed to be present and should never be touched by the user. They're just plugins to keep the structure clean.

splinter98:
I have thought about this (and similar related) issues a fair amount over the years in my spare time. 

The main issue I have with moving everything to PyPi is the core of Vapoursynth isn't Python. Only the (supported) front end.

Bar a few help modules all plugins are C++ and wouldn't package well inside python. Moving Vapoursynth to PyPi would require restructuring the whole VS ecosystem, which while I'm not adverse to, in this case I just can't see it being beneficial. For example when using VS/AVFS or VSEditor or other third party utilities.

[...]

PR #320, opened 2017-07-06
Improve setup.py to use system vapoursynth.dll

stuxcrystal:
This Pull-Request improves the setup.py file to allow it to install via pip.

# How does it accomplish this?
This pull-request actually solves a related problem: Allowing the setup.py to detect the vapoursynth.dll installed by the installer.

The setup.py routine now try to locate the vapoursynth.dll using three ways:
1. vapoursynth.dll and vapoursynth.lib from `Release` or `x64/Release`
2. vapoursynth.dll and vapoursynth.lib from PATH
3. vapoursynth.dll and vapoursynth.lib located in the Registry

The located vapoursynth.dll-file will be copied to Lib\\site-packages during the installation and shipped in binary distributions of this package.

On linux there are no changes and no additional files except the module (and meta-data about the package) itself will be copied into site-packages. Linux based binary-distributions will not contain a vapoursynth.so file.

# Dependencies
This PR introduces a dependency to setuptools. Since setuptools is packaged with current python versions, this will likely not prove to be a problem.

Installing from source needs Visual Studio 2015. Alternatively you can install vapoursynth inside the Native VS2017 build shell of the desired architecture.

# Python distributions
The setup.py allows binary and source distributions to be compiled and uploaded to pypi. This section will address the state of both commands.

## Source distributions `sdist`
Using `python setup.py sdist`, a source distribution will be created that will create a `.tar.gz` (or using `--format=zip` a zip-file) containing the neccessary files to install the vapoursynth package by compiling.

This package will neither ship with a vapoursynth.lib nor with a vapoursynth.dll or any other compiled file.

## Wheels `bdist_wheel`
Using `python setup.py bdist_wheel`, you can create a binary distribution that can be installed without the compiling anything.

The wheel will ship with their own compiled version of `vapoursynth.dll`

Since no compilation is being done here, the pip-bug below does not apply when installing vapoursynth using a wheel.

## Instalation path
The setup will install both the vapoursynth.dll as well as the compiled cython module directly into site-packages. `pip` will remove both files when updating or removing the package.

## Release considerations
The `setup.py` file now contains the current release-number as its _major_ version. This allows packages dependent on VapourSynth to specify the minimum required version of vapoursynth inside their setup.py file. Make sure the version remains up-to-date to allow users to update and specify their vapoursynth installation using `pip`

Additionaly, the source distribution for vapoursynth as well as binary distributions for Windows x64, Windows x32, Linux x64 and Linux x32 should be uploaded to pypi.

# Where to go from here.
Since we can expect pip to be installed on all Windows systems, we can change the module-installation routine in the Windows installer to instead install a binary-wheel created by setup.py using the pip installation of the package. The binary-wheel can be included by the installer.

# Known Problems
Python 3.6 changed the way I/O is handled on Windows causing a bug on non-english Windows installations which is fixed here https://github.com/pypa/pip/pull/4486
Until this fix is shipped with a new release of pip,  change the code-page for the current command shell to UTF-8 by executing `chcp 65001`.  This is not neccessary when installing using a binary distribution.

# Afterword
Since I cannot foresee all build-environments this thing is possibly gonna run on, additional testing may be helpful.

# Additional things added in this PR
Added zimg and AviSynth+ to the .gitignore-file

# References to other issues
Implements #283

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,7 @@ ltmain.sh
 missing
 vspipe
 *.pc
+
+dist/
+.eggs
+*.egg-info/
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,6 @@ vspipe
 dist/
 .eggs
 *.egg-info/
+
+zimg/
+AviSynthPlus
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,3 +4,5 @@ recursive-include test *
 recursive-include src/cython *.pyx *.pxd
 recursive-include include *.h *.asm
 recursive-exclude include/cython *
+
+global-exclude *.dll
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,6 @@
+include COPYING.LGPLv2.1
+
+recursive-include test *
+recursive-include src/cython *.pyx *.pxd
+recursive-include include *.h *.asm
+recursive-exclude include/cython *
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,9 @@ if is_win:
 
         return value
 
-    library_dirs.append(join(query(winreg.HKEY_LOCAL_MACHINE, REGISTRY_PATH, REGISTRY_KEY), "sdk", lib_suffix))
+    base_dir = query(winreg.HKEY_LOCAL_MACHINE, REGISTRY_PATH, REGISTRY_KEY)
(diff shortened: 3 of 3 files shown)


myrsloik:
Interesting, especially since the python module rarely changes nowadays. I'll probably merge it in a day or two when I'm not too sleepy to review the code

PR #322, opened 2017-07-08
Add easier way to register new formats.

stuxcrystal:
I thought, it would be awesome to have a method that can create a new Format based on a preexisting format.

Addtionionally:
* Implement `__int__` on `Format` to allow Format-objects to be passed into filters.


dubhater:
Nice.

PR #326, opened 2017-07-22
Update vsresize to support AVX-512

sekrit-twc:


myrsloik:
Would it be possible to make the alignment only be 64 bytes if the cpu supports avx512? It feels kinda iffy to require 64 byte alignment on arm and other cpu types as well and simply waste more memory (the original 32 bytes everywhere is probably a mistake in some ways but meh)

sekrit-twc:
It is possible, but the most common frame dimensions are already 64-byte aligned:

320
640
720 *not this one*
768
1024
1280
1920
...

PR #334, opened 2017-09-12
Remove limit on LibASS for Karaoke FX

stuxcrystal:
In a land far far away in a Kingdom ruled by Queen Merkel there was a drug called karaoke-fx. Using equipment like NyuFX people could create this drug. However, like all proper drugs, its effects diminish with prolonged use and as such needed more and more of this drug. This commit allows to enter ASS-Files larger than 50MiB.

A program should not assume sane users.


dubhater:
> A program should not assume sane users.

That was kind of the point of that limit. :P

PR #335, opened 2017-09-29
Levels: Fix interpretation of gamma

Frechdachs:
This bug only affected integer sample type.


myrsloik:
Well spotted, a variation on this haa been committed

PR #346, opened 2017-12-13
Fix link to znedi3

oyvindln:
The original link went to  `nnedi3cl` instead, so I added a separate entry for that and linked `znedi3` to the correct github page.


myrsloik:
Applied a similar fix without the duplicated nnedi3cl entry

oyvindln:
Ah I somehow didn't notice that it was already linked.

PR #353, opened 2018-01-11
Add missing stddef.h includes for ptrdiff_t

wm4:
Otherwise fails to build.

---

Not sure why, maybe a newer gcc or glibc with cleaner headers?


wm4:
Aw, should have looked at the open PRs - I'm duplicating #352.

PR #362, opened 2018-01-24
Better reflection and improved compatibility with older VS-Wrapper versions regarding outputs.

stuxcrystal:
I think it is a bad idea to always store `(clip, alpha)`-tuples, even if `alpha` is not passed to `set_output`.

This is a breaking change and I think we make it compatible again.

### DIFF
--- a/doc/pythonreference.rst
+++ b/doc/pythonreference.rst
@@ -188,7 +188,7 @@ Classes and Functions
 .. py:function:: get_output([index = 0])
 
    Get a previously set output node. Throws an error if the index hasn't been
-   set.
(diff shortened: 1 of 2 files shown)


myrsloik:
Wouldn't calling it alphaoutputtuple make more sense?

stuxcrystal:
Actually, it would.

dubhater:
```python
clip.set_output(0, alpha=alphaclip)
clip = vs.get_output(0)
isinstance(clip, tuple)
```
This is no longer true, is it?

myrsloik:
Indeed, it probably broke all the other places.

stuxcrystal:
@dubhater In this particular case, it is true.

However, it is not guaranteed anymore that the output it always a tuple. This is what broke the code.

PR #372, opened 2018-02-19
Add support for a special _vapoursynth_-Attribute

stuxcrystal:
Note: **Don't merge yet.** This pull-request is here for getting more input and feedback.

This code would allow your own objects to be passed into a vapoursynth function. If it has a `_vapoursynth_`-Attribute, its value will be used for all further processing.

### DIFF
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -471,6 +471,8 @@ cdef void dictToMap(dict ndict, VSMap *inm, VSCore *core, const VSAPI *funcs) ex
                 val = [val]     
 
         for v in val:
+            if hasattr(v, '_vapoursynth_'):
+                v = v._vapoursynth_
             if isinstance(v, VideoNode):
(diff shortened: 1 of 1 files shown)


myrsloik:
Why would I want to do that?

stuxcrystal:
Well the idea is that someone could potentially write a class that does things and can be passed directly into the argumetns of a filter.

```py
class MyExprGenerator(object):
    # ... Lots of stuff that creates your expr
    @property
    def _vapoursynth_(self):
        return "somethings that generates your expr."

# ... A lot more stuff...
core.std.Expr([my_clip], [MyExprGenerator.X / 2])
```

BluBb-mADe:
Wouldn't it be easier to just check if the object is callable and call it instead of this whole _vapoursynth_ business?

stuxcrystal:
It is kinda the default way the Python STDLib does it. Also: It checks for callable-objects because you need to send callback-functions. So if the object is callable it is wrapped into a VapourSynth function.

PR #380, opened 2018-04-14
Add setup.py for portable installations

stuxcrystal:
# What is this about?
The current setup.py only installs the `vapoursynth.pyd` and `VapourSynth.dll` neccessary for working with the pre-existing VapourSynth installation.

This PR adds a setup.py for portable installations, copying the neccessary files for portable installations into the Python-directory.

# What do I need to know
This setup.py **will** require the matching vapoursynth-installation installed via `setup.py` (of VapourSynth itself) or via `PyPI`. The installation of this package will attempt to redownload and install the package prior to installation of the portable edition if it hasn't been installed beforehand.

As such, the source-distribution and binary-wheels will not include the `VapourSynth.dll` itself as it is included in its requirement.

# Where do the files come from?
There are essentially two modes of this `setup.py`-file:
1. It finds the directory `buildp64` (or `buildp32` on 32bit-Systems) created by `make portable.bat` and uses its directory structure to copy the files to the correct destination.
2. It finds the file `VapourSynth.dll` and assumes it was included in the portable .7z-file and takes the files from the extracted portable directory.

# What is included?
Currently, all plugins provded in `vapoursynth(32|64)/coreplugins` are included as well as the files required to run `VSPipe.exe`.

`VapourSynth` itself is not included.

### DIFF
--- a/installer/MANIFEST.in
+++ b/installer/MANIFEST.in
@@ -1,0 +1,2 @@
+global-include *.dll portable.vs .keep VSPipe.exe
+global-exclude VapourSynth.dll
--- a/installer/make portable.bat
+++ b/installer/make portable.bat
@@ -1,4 +1,4 @@
-rem extract version string
+em extract version string
 for /F "tokens=2 delims='" %%a in ('findstr /C:"#define Version" vsinstaller.iss') do set v=%%a
 @echo %v%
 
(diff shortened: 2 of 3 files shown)


stuxcrystal:
Any chance I get a comment on this PR?

myrsloik:
Oooook. So basically it makes it convenient to install portable VS into a per user python install (with no additional permissions needed)?
I guess it's an acceptable addition, at least until I figure out a way to have a sane permissionless installer.

stuxcrystal:
That is true.

You can install the portable edition from the 7z-archive you publish in your Releases by executing this command from within the extracted archive:
```cmd
py -3.6 pip install .
```

or even easier by doing
```cmd
py -3.6 -m pip install vapoursynth-portable
```

The latter one already works. :)

PR #385, opened 2018-04-26
Don't skip resizing if subpixel shifting is required.

sekrit-twc:


src/core/vsresize.cpp
```diff
@@ -440,6 +440,15 @@ bool operator!=(const zimg_image_format &a, const zimg_image_format &b) {
     return !operator==(a, b);
 }
 
+bool is_shifted(const zimg_image_format &fmt) {
+    bool ret = true;
+    ret = ret && !std::isnan(fmt.active_region.left) && fmt.active_region.left != 0;
+    ret = ret && !std::isnan(fmt.active_region.top) && fmt.active_region.top != 0;
```
HolyWu: What if the user needs to shift only one direction, but not both?

sekrit-twc: I reversed the logic.

PR #403, opened 2018-07-12
Switch to python 3.7 on windows

fdar0536:


myrsloik:
Rejected, Python 3.7 will be the default for R45. Also don't make patches where every line is changed for no reason.

PR #465, opened 2019-07-30
Be more friendly to Python autocompletion engines

Endilll:
#### Problem
Microsoft's engine doesn't list any plugins at all for core object (`_CoreProxy`). Jedi list the plugins, but doesn't list functions.
#### Reason
Microsoft's engines examines objects in a static way via dict of the type. Jedi does the same, but also calls type's `__dir__()` — that's why it's able to list plugins. No one triggers `__getattr__()` or `__getattribute__()` to get actual object for given plugin to list it's attributes. This behavior is intended since engines aims to execute less code. 
#### Solution
Fetch all the plugins objects from core and add them to `_CoreProxy` as if they were attributes defined in source. Metaclasses is the right mechanism for that.
#### Result
Both autocompletion engines lists the plugins, and Jedi lists functions inside plugins.
#### Testing code
```
import jedi
from vapoursynth import core

script = jedi.Interpreter('core.', [locals()])
print(script.completions())
```

### DIFF
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -1754,7 +1754,8 @@ cdef class Plugin(object):
         return attrs
 
 cdef Plugin createPlugin(VSPlugin *plugin, str namespace, const VSAPI *funcs, Core core):
-    cdef Plugin instance = Plugin.__new__(Plugin)
+    t = type('Plugin', (Plugin,), {})
+    cdef Plugin instance = t.__new__(t)
     instance.core = core
     instance.plugin = plugin
     instance.funcs = funcs
@@ -1765,6 +1766,14 @@ cdef Plugin createPlugin(VSPlugin *plugin, str namespace, const VSAPI *funcs, Co
         if plugin_dict['namespace'] == namespace:
             instance.__doc__ = plugin_dict['name']
             break
+
+    cdef VSMap *m = funcs.getFunctions(plugin)
+    for i in range(instance.funcs.propNumKeys(m)):
+        cname = instance.funcs.propGetKey(m, i)
+        name = cname.decode('utf-8')
+        signature = instance.funcs.propGetData(m, cname, 0, NULL).decode('utf-8').split(';', 1)
+        setattr(type(instance), name, createFunction(name, signature[1], instance, instance.funcs))
+    funcs.freeMap(m)
 
     return instance
 
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -1655,8 +1655,16 @@ cdef Core vsscript_get_core_internal(int environment_id):
         _cores[environment_id] = createCore()
     return _cores[environment_id]
     
-cdef class _CoreProxy(object):
+class CoreMeta(type):
+    def __new__(cls, name, bases, dct):
+        core = get_core()
+        for plugin in core.get_plugins().values():
+            name = plugin['namespace']
+            dct[name] = core.__getattr__(name)
 
+        return super(CoreMeta, cls).__new__(cls, name, bases, dct)
+
+class _CoreProxy(metaclass=CoreMeta):
     def __init__(self):
         raise Error('Class cannot be instantiated directly')
     
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -1655,16 +1655,7 @@ cdef Core vsscript_get_core_internal(int environment_id):
         _cores[environment_id] = createCore()
     return _cores[environment_id]
     
-class CoreMeta(type):
-    def __new__(cls, name, bases, dct):
-        core = get_core()
-        for plugin in core.get_plugins().values():
-            name = plugin['namespace']
-            dct[name] = core.__getattr__(name)
-
-        return super(CoreMeta, cls).__new__(cls, name, bases, dct)
-
-class _CoreProxy(metaclass=CoreMeta):
+class _CoreProxyVSScript(object):
     def __init__(self):
         raise Error('Class cannot be instantiated directly')
     
@@ -1685,7 +1676,16 @@ class _CoreProxy(metaclass=CoreMeta):
     def __setattr__(self, name, value):
         setattr(self.core, name, value)
     
-core = _CoreProxy.__new__(_CoreProxy)
+def _make_standalone_proxy():
+    core = get_core()
+    dct = {"core": core}
+    for name in dir(core):
+        if name.startswith("__") and name.endswith("__"):
+            continue
+        dct[name] = getattr(core, name)
+    return type('_CoreProxyStandalone', (), dct)
+_CoreProxyStandalone = _make_standalone_proxy()
+core = _CoreProxyStandalone()    
     
 
 cdef class Plugin(object):
@@ -2150,4 +2150,5 @@ cdef public api int vpy_initVSScript() nogil:
             return 1
         global _using_vsscript
         _using_vsscript = True
+        core.__class__ = _CoreProxyVSScript
         return 0
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -1783,12 +1783,6 @@ cdef class Function(object):
     cdef readonly str name
     cdef readonly str signature
     
-    @property
-    def __signature__(self):
-        if typing is None:
-            return None
-        return construct_signature(self.signature, injected=self.plugin.injected_arg)
-
     def __init__(self):
         raise Error('Class cannot be instantiated directly')
 
@@ -1869,7 +1863,9 @@ cdef class Function(object):
         return retdict
 
 cdef Function createFunction(str name, str signature, Plugin plugin, const VSAPI *funcs):
-    cdef Function instance = Function.__new__(Function)
+    dct = {'__signature__': construct_signature(signature, injected=plugin.injected_arg)}
+    t = type('Function', (Function,), dct)
+    cdef Function instance = t.__new__(t)
     instance.name = name
     instance.signature = signature
     instance.plugin = plugin
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -36,6 +36,7 @@ from types import MappingProxyType
 from collections import namedtuple
 from collections.abc import Iterable, Mapping
 from fractions import Fraction
+from warnings import warn
 
 # Ensure that the import doesn't fail
 # if typing is not available on the python installation.
@@ -1868,7 +1869,7 @@ cdef Function createFunction(str name, str signature, Plugin plugin, const VSAPI
         try:
             dct['__signature__'] = construct_signature(signature, injected=plugin.injected_arg)
         except ValueError as e:
-            print("Constructing signature for {}.{}() failed: {}".format(plugin.namespace, name, str(e)))
+            warn('Constructing signature for {}.{}() failed: {}'.format(plugin.namespace, name, str(e)))
     t = type('Function', (Function,), dct)
     cdef Function instance = t.__new__(t)
     instance.name = name
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -1867,10 +1867,16 @@ cdef Function createFunction(str name, str signature, Plugin plugin, const VSAPI
     dct = {'__signature__': None}
     if typing is not None:
         try:
-            dct['__signature__'] = construct_signature(signature, injected=plugin.injected_arg)
+            sign = construct_signature(signature, injected=plugin.injected_arg)
+            annotations = dict(sign.parameters)
+            annotations['return'] = sign.return_annotation
+
+            dct['__signature__'] = sign
+            dct['__annotations__'] = annotations
         except ValueError as e:
             warn('Constructing signature for {}.{}() failed: {}'.format(plugin.namespace, name, str(e)))
     t = type('Function', (Function,), dct)
+
     cdef Function instance = t.__new__(t)
     instance.name = name
     instance.signature = signature
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -189,7 +189,7 @@ def _construct_parameter(signature):
     elif type == "float":
         type = float
     elif type == "data":
-        type = typing.Union[str, bytes, bytearray]
+        type = typing.AnyStr
     else:
         raise ValueError("Couldn't determine type")
 
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -1865,7 +1865,10 @@ cdef class Function(object):
 cdef Function createFunction(str name, str signature, Plugin plugin, const VSAPI *funcs):
     dct = {'__signature__': None}
     if typing is not None:
-        dct['__signature__'] = construct_signature(signature, injected=plugin.injected_arg)
+        try:
+            dct['__signature__'] = construct_signature(signature, injected=plugin.injected_arg)
+        except ValueError as e:
+            print("Constructing signature for {}.{}() failed: {}".format(plugin.namespace, name, str(e)))
     t = type('Function', (Function,), dct)
     cdef Function instance = t.__new__(t)
     instance.name = name
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -1863,7 +1863,9 @@ cdef class Function(object):
         return retdict
 
 cdef Function createFunction(str name, str signature, Plugin plugin, const VSAPI *funcs):
-    dct = {'__signature__': construct_signature(signature, injected=plugin.injected_arg)}
+    dct = {'__signature__': None}
+    if typing is not None:
+        dct['__signature__'] = construct_signature(signature, injected=plugin.injected_arg)
     t = type('Function', (Function,), dct)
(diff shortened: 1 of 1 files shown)


myrsloik:
How does this interact with plugin loading? Or this will only reveal autoloaded plugins?

Endilll:
Autoloaded only

myrsloik:
There's also the option of using less getattr trickery and make it possible for the core to notify python whenever a new plugin function is added. Then you could turn things into proper objects with. That'd solve the problems for everything, right?

Endilll:
Not sure if I grasped your idea, but for me it seems like you suggest to add now-missing `__dict__` attribute to `Core` or `_CoreProxy` to store dynamic attributes, and add plugin objects there upon receiving notification about loading new plugin coming from vapoursynth.dll via C API.

While this approach is definitely better than `__getattr__()`, actual behavior of the engines should be checked. I have my doubts about Microsoft's engine since it's pretty conservative.

stuxcrystal:
I doubt that it is as easy as it seems.

In Python we have two classes:
- `Core` which actually handles plugins and stuff for you. You can easily add attributes to this class (or dynamically generated subclass in this case, if I understand you correctly) 
- `_CoreProxy` however is a different beast. It describes a Singleton whose attributes are when using vsscript dependent on the thread it is accessed from.

The core problem lies in this proxy object. As the contents of the dict must be sensitive to thread-locals. The `__dict__` object must always be backed and will always be downgraded to a `dict` if a subclass of `dict` is assigned to `__dict__` or passed as a parameter of `type` during class creation.

This means there is no clean way I can think of to support both the Ms Completion engine and vsscript introspection.

I'd vote fore maintaining compatibility with previous versions and keep Core Proxy as is.

Note:
If vsscript is enabled it is imperative to make sure jedi is called from a thread where the `Core` (or it's proxy) you want to introspection is active.

Endilll:
Thanks for pointing out on VSScript issues. I haven't considered it when I worked on PR, and I agree that if merged like this, my PR breaks core isolation into shatters.

As for introspections while in VSScript, I don't think we need this at all since using VSScript API implies that you're not an ordinary VS user who would benefit much from autocompleting of plugin functions names.

@myrsloik as for your idea about dynamic updates, Jedi doesn't search attributes in `__dict__` of the instance, but it does search in the `__dict__` of the type (more on that below. And I'd expect that this won't go together with caching that engines most probably uses.

#### How Jedi is trying to fetch plugin's functions (PR not applied)
Using `core.eedi3.` (mind trailing period) as example:
1. `CoreMeta.__new__(name, bases, dct)` — the obvious first step is to evaluate class definitions. In out case `name` is `_CoreProxy`.
2. Retrieve core object and see that it's type is `_CoreProxy`.
3. `_CoreProxy.__getattribute__(__class__)` — retrieving `_CoreProxy` class itself.
4. `CoreMeta.__getattribute__(__dict__)` — retrieving class's dict (normally the same as `dct` from step 1).
5. Look for `eedi3` in `dct` and find nothing.
6. 'No suggestions' since object for `eedi3` is not found.

`CoreMeta` injects plugin objects on step 1.

Steps 3-5 are called `getattr_static` in Jedi sources, and this is the only routine that retrieves objects. If it fails, `__dir__()` is used to get at least names of the attributes without any further lookup.

#### Default attribute lookup (simplified for our case; detailed [here](https://blog.ionelmc.ro/2015/02/09/understanding-python-metaclasses/#object-attribute-lookup))
1. `__getattribute__()`
2. `__dict__` of the instance
3. `__dict__` of the type
4. `__getattr__()`

#### Suggestions
I see two approaches that could be taken, aside from leaving everything as it is:
1. Expose to regular VS users object of `Core` type itself or another proxy, and use metaclass on it, leaving all the complexity involving multiple core instances to current `_CoreProxy`. This is the best solution from my point of view, but I bet there're pitfalls I'm not aware about.
2. Make existing getattr machinery even more complicated hooking step 3 of Jedi's lookup and injecting plugins into class's dict. Just this, without any further details, makes me think that this is too hacky.

stuxcrystal:
Note that most if not all vapoursynth-aware editors use vsscript in some way. And I know at least one that does use Jedi to perform autocompletions (Yuuno). 

Other than that: I can't wrap my head around your suggestion 1, can you explain it in more detail?

Endilll:
First suggestion was about leaving current `_CoreProxy` class for VSScript, and introducing new proxy type that has static core, which is suitable for people who write scripts and doesn't need VSScript (if they know about it existence). Since it'll have static core by design, we can freely use metaclass I wrote to help engines get plugin objects.

Meanwhile, I took a closer look at `_CoreProxy` and come up with a simpler idea: in addition to metaclass provide custom `__getattribute__()`, which would bypass class's dict (which is populated with plugins from wrong core). That way engines'll get their plugin objects, and everyone else's attribute lookup'll will be forwarded to appropriate core. I can implement this, if you agree with the approach.

By the way, is autocompletion in Yuuno not able to list plugin functions, too?

stuxcrystal:
You said that Jedi never looks at the `__dict__` of the instance, right?

What about we switch out the `__class__`-attribute when vsscript is being activated.

What I mean is: We implement two classes (that must share the same meta-class), let's call the first one `_CoreProxyStandalone` and the second one `_CoreProxyVSScript`.

The first one has all plugins in `__dict__` the other one does not. 

When vsscript is being activated, `core.__class__` is overriden: from `_CoreProxyStandalone` to `_CoreProxyVSScript`

P. S. About Meta-Classes: Why exactly do you think meta-classes are needed again? Can't you just dynamically create new classes using `type(name, bases, dict) `. No sub-classing of `type` needed.

Endilll:
Sounds good. Actually, `type()` can save us from defining second proxy class. I think implementation should be like this:
```
class _CoreProxy:
    def __getattribute__(self, name):
        if name != '__class__':
            return super().__getattribute__(name)
        
        global _using_vsscript
        if _using_vsscript:
            return super().__getattribute__('__class__')

        # _CoreProxyStandalone should be generated once and stored, I think.
        # Or not, and that would open the way to plugin list updates,
        # but that would be pretty slow, I suppose. 

        core = super().__getattribute__('core')
        dct = {}
        for plugin in core.get_plugins().values():
            name = plugin['namespace']
            dct[name] = getattr(core, name)
        return type('_CoreProxyStandalone', (), dct)
```

stuxcrystal:
I personally have a aversion from using `__getattribute__` as it is called on all attribute accesses. Especially when trying to monkey-patch magic-methods.

I prefer `__getattr__` as it has lower priority than normal attribute lookup and thus lets instances more or less act like normal python instances.

Monkey-patching `__class__` would allow the two modes (vsscript and standalone) have different places to implement their attribute-lookups separately. I think this is a cleaner solution.

And if you add a notification algorithm, `_CoreProxyStandalone` doesn't need to use `__getattr__` at all.

On the python side, I'd let the core notify the wrapper-class. Which then notifies the core-proxy (`_CoreProxyVSScript` would be a noop. This allows other programs to implement their own for GUIs, etc.). The standalone implementation just adds the plugin to the class' `__dict__`.

Endilll:
Replacing type object the way code above does seems too hacky, because plugin lookup fails.

Your solution isn't that clear to me to implement and test it right away, especially the part about overriding `__class__`. Where this would be done? But I'm willing to participate in its implementation nevertheless.

Side note: `__getattribute__()` generally doesn't participate in implicit special method lookup (see [docs](https://docs.python.org/3/reference/datamodel.html#special-method-lookup))

stuxcrystal:
It would be done here:

https://github.com/vapoursynth/vapoursynth/blob/a22fe2578fb81bb3bb735239dd5c135ee02b7664/src/cython/vapoursynth.pyx#L2143

Literally the last function of the file.

Why would plugin lookup fail?

Endilll:
Probably `instance.__getattribute__('__class__')` isn't the only way to get type object of the instance. Jedi gets real type object somehow, compares replaced one to it, and ignores it or whatever. But this is just a speculation.

stuxcrystal:
Ehm... There is always `type(obj: Any) -> Type` to get the class-object. Which, incidentally, is what jedi uses.

```python
>>> import jedi
>>> class C:
...     x = (lambda self: None)
...
>>> class D:
...     y = (lambda self: None)
...
>>> y = C()
>>> script = jedi.Interpreter('y.', [locals()])
>>> print(script.completions())
[<Completion: x>, ...]
>>> y.__class__ = D
>>> script = jedi.Interpreter('y.', [locals()])
>>> print(script.completions())
[<Completion: y>, ...]
>>> y.__class__ = C
>>> print(script.completions())
[<Completion: x>, ...]
```

Doing that seems to work even when reusing interpreter instances.

In the end, I propose something along the lines of:

```py
def _make_standalone():
        core = get_core()
        dct = {"core": core}
        for name in dir(core):
            if name.startswith("__") and name.endswith("__"):
                continue
            dct[name] = getattr(core, name)
        return type('_CoreProxyStandalone', (), dct)
_CoreProxyStandalone = _make_standalone()
core = _CoreProxyStandalone

class _CoreProxyVSScript:
    @property
    def core(self):
        return get_core()

    def __getattr__(self, attrname):
        return getattr(self.core, attrname)

def vpy_initVSScript():
     # ...
     core.__class__ = _CoreProxyVSScript
     # ...
```

Endilll:
Now I got it. If it works, I don't have anything to add.
If not, I'd blame `core.__class__ = _CoreProxyVSScript` line.

Endilll:
Not like it matters now, but what I was talking about earlier when I said that plugin lookup fails is better depicted in the following snippet:
```
class B:
    def __getattribute__(self, name):
        if name != '__class__':
            return super().__getattribute__(name)
        return type('C', (), {})

b = B()
print(type(b))      # <class '__main__.B'>
print(b.__class__)  # <class '__main__.C'>
```
_Tested on CPython 3.7.4_

Endilll:
It makes me wonder if the same technique can be applied to provide function signatures. But there's a pair of pointers inside, so it can't be simply converted to regular Python class.

stuxcrystal:
Well... Right now, this works well enough.

We could use the `__new__`-class-method to do provide better function signatures.

Either [here](https://github.com/vapoursynth/vapoursynth/blob/79b571bf42aad690bac73dc18cb76317b1f155f6/src/cython/vapoursynth.pyx#L1771):
```python
class Function:
    def __new__(cls, ...):
        cls = type(name, (cls, ), {"__signature__": cls.construct_signature(...)})
        return cls(...)
```

or [here](https://github.com/vapoursynth/vapoursynth/blob/79b571bf42aad690bac73dc18cb76317b1f155f6/src/cython/vapoursynth.pyx#L1862):
```python
def createFunction(...):
    cls = type(name, (cls, ), {"__signature__": cls.construct_signature(...)})
    function = cls.__new__(...)
```

Endilll:
Oh, my previous comment was a bit misleading. I was talking about `Plugin` providing `Function`s via `__getattr__()`, which doesn't go well with autocompletion.

stuxcrystal:
The same things still apply. Its called `Plugin` and `createPlugin` respectively.

Endilll:
Not much left to be desired at this point looking from Visual Studio Code UX side, improving which was my original intent. I plan to left nothing to be desired, but that'll be a subject of another PR at some point in the future.

Endilll:
Just to check: you expecting something from my side or your hands are just full with other stuff right now?

Endilll:
Meanwhile, I found out that if plugin provides invalid signature for function, exception text is anything but clear for user:
```
Traceback (most recent call last):
  File "test\completion.py", line 1, in <module>
    import vapoursynth as vs
  File "src\cython\vapoursynth.pyx", line 1801, in init vapoursynth
    _CoreProxyStandalone = _make_standalone_proxy()
  File "src\cython\vapoursynth.pyx", line 1799, in vapoursynth._make_standalone_proxy
    dct[name] = getattr(core, name)
  File "src\cython\vapoursynth.pyx", line 1599, in vapoursynth.Core.__getattr__
    return createPlugin(plugin, name, self.funcs, self)
  File "src\cython\vapoursynth.pyx", line 1889, in vapoursynth.createPlugin
    setattr(type(instance), name, createFunction(name, signature[1], instance, instance.funcs))
  File "src\cython\vapoursynth.pyx", line 1983, in vapoursynth.createFunction
    t = type('Function', (Function,), dct)
  File "src\cython\vapoursynth.pyx", line 228, in vapoursynth.construct_signature
    return inspect.Signature(tuple(params), return_annotation=vapoursynth.VideoNode)
  File "C:\Program Files\Python37\lib\inspect.py", line 2788, in __init__
    raise ValueError(msg)
ValueError: non-default argument follows default argument
```
I got this on [vcfreq](http://www.avisynth.nl/users/vcmohan/vcfreq/vcfreq64.7z).
Since this PR effectively enables signature checking for all plugins on importing VS, I think we should provide more helpful message for this case. I haven't found any other warnings about non-fatal errors in vapoursynth.pyx, so I went with `print()`.

src/cython/vapoursynth.pyx
```diff
@@ -1865,7 +1865,10 @@ cdef class Function(object):
 cdef Function createFunction(str name, str signature, Plugin plugin, const VSAPI *funcs):
     dct = {'__signature__': None}
     if typing is not None:
-        dct['__signature__'] = construct_signature(signature, injected=plugin.injected_arg)
+        try:
+            dct['__signature__'] = construct_signature(signature, injected=plugin.injected_arg)
+        except ValueError as e:
+            print("Constructing signature for {}.{}() failed: {}".format(plugin.namespace, name, str(e)))
```
stuxcrystal: Use `warnings` instead of `print`. https://docs.python.org/3/library/warnings.html
Warnings can be explicitly captured. Print-calls can't.

stuxcrystal:
Also I found out why this happens:
At https://github.com/vapoursynth/vapoursynth/blob/56b313f925e737733d0b0cbb99e08fa403421199/src/cython/vapoursynth.pyx#L1869 the parameter `injected` is always `None` which means the first parameter is never removed.

Only at
https://github.com/vapoursynth/vapoursynth/blob/56b313f925e737733d0b0cbb99e08fa403421199/src/cython/vapoursynth.pyx#L1221 the parameter is set. Which is after the signature has been created.

Endilll:
I don't get how not removing first parameter makes signature invalid. Marking first argument as `opt` with following non-opt?

In case above cause was different: last argument wasn't marked as `opt`:
```
clip:clip;frad:int:opt;test:int:opt;morph:int:opt;gamma:float:opt;fspec:int[];
```

stuxcrystal:
Could it be that I found another bug?

Is the signature for `core.std.BlankClip().std.ShufflePlanes` (or any other function for that matter) correct?

Endilll:
Now I see what are you getting at. Right now autocompletion doesn't recognize any types in signature, so it fails at the third dot. I plan to work on it, but later (I don't want to block this PR).

You think we should try to remove first argument on the fly when injected argument is present and hope that jedi would recognize this?

stuxcrystal:
What's the state of this PR?

Endilll:
I've done all I wanted for now, and don't plan to return to this topic anytime soon, so I believe it's ready to be reviewed and merged.

Endilll:
Jedi not recognizing any types in signature turned out to be a bug on it's side. [Fix](https://github.com/davidhalter/jedi/pull/1431) is merged.

Endilll:
@stuxcrystal I think of replacing this annotation with `typing.AnyStr`, which covers `str` and `bytes`, but not `bytearray`. Is it fine or I'm missing something important?
https://github.com/vapoursynth/vapoursynth/blob/66d7d04399bfaa5b2054ec0aa198f99c4bfc330a/src/cython/vapoursynth.pyx#L191

I'm doing this in order to reduce noise in function signatures that are shown to users when cursor is inside braces of function call.

stuxcrystal:
You can try, but type-annotations are also parsed by static-code-analyzers thus showing warnings.
On the other hand, I'm not sure anyone uses VapourSynth with `bytearrays`.

intro94:
Are you going to merge this? I applied these changes to myself — it is very convenient! But due to the fact that you did not get the pull request, I have to rebuild VS frome source code each time you update VS. :(

myrsloik:
1. I really don't have the python knowledge to review this well
2. If it does happen it'll be after audio support

Endilll:
A status update for out-of-box autocompletion support:
* [CPython PR](https://github.com/python/cpython/pull/19714) got merged
* [Jedi PR](https://github.com/davidhalter/jedi/pull/1564) got merged

Combined with this PR, that's how function signature looks like:
`clip: vapoursynth.VideoNode, width: Optional[int]=None`

I think that this is the best result one can demand from upstream tools without special handling for vapoursynth, and pretty good overall.

myrsloik:
Do you want to revive this in the doodle1 branch? Audio support is now done as advertised

Endilll:
Sure. But I don't feel confident resolving the conflict in cython code by myself.

Endilll:
In your opinion, is there a value in this PR given that stubs generator was added to vsrepo in the meantime? Looks like it doesn't handle autoloaded plugins very well if they are copied by hand, but is this use case still important? Maybe there're others I'm not aware of.

myrsloik:
> In your opinion, is there a value in this PR given that stubs generator was added to vsrepo in the meantime? Looks like it doesn't handle autoloaded plugins very well if they are copied by hand, but is this use case still important? Maybe there're others I'm not aware of.

You have a point. Maybe the vsrepo stubs already solved things. Only crazy people gather up their plugin dirs manually.

Issue #324, opened 2017-07-19
Introduce wait-state for filters


stuxcrystal:
This is just an idea. I will not take it personally if it is rejected. Discussion on this proposal is encouraged.

# Explanation
In essence, we allow the filter to signal that it is waiting for an external event and that the core can nowuse the worker to process other frames.

# Proposed API

A simple way to implement this is via a similar method to `requestFrameFilter`. It can be called from the getFrame-function of the filter to signal that the frame will enter a wait-state after exiting the function. The core will return a pointer to a handle-object that can be used to identify which filter to wake up again. I will call this function "setFrameWaiting()". On failure, this function should return `NULL`.

Waking up a filter will require another new API-function. I will call this function `notifyFrame`. It takes a single argument; the pointer to the handle that has been returned. It will return nonzero value on failure.

After notify frame is called, the core will schedule the getFrame-call with a new ActivationReason `arNotified`.

Note that the filter may still receive frames commisioned by calls to `requestFrameFilter()`. However my proposal would forbid calling `setFrameWaiting()` while the filter is in the state "sleeping".

Should the frame be notified while a getFrame-call to the same instance is currently being executed, the core will wait for the previous getFrame call to finish before invoking it again with arFilterNotified.

## Thread Safety
`notifyFrame` need to be thread safe.

## API Additions
* `void *setFrameWaiting(VSFrameContext *frameCtx)`
* `int notifyFrame(void *handle)`
* `arNotified`

## Overview:

    core                          frame      
     |
     | = getFrame(arInitial) =======> | 
     | <===== setFrameWaiting() ===== | 
     | -------------- void *handle- > |
     |                                |
     .                                .
     .                                .
     .                                .                  external Thread
     |                                |                         |
     | <=== notifyFrame(handle) =============================== |
     | -------------------------------------------------------> |
     |                                |                         |
     | = getFrame(arFrameNotified) => |                         |
     | <- frame --------------------- |                         |
     |                                |                         |

# Use-Cases

1. Implementation of I/O-operations. I/O-operations usuall don't require CPU. Therefore the core may use the resources more efficiently if it could schedule another frame while I/O-operations are in process.
   An example implementation would look like this: The `getFrame`-method may call `setFrameWaiting` and passing the returned handle to another thread. When the operation completes this thread will call `notifyFrame` with the handle passed to the thread.
2. This model can be easily used to interoperate with other concurrency models.

BluBb-mADe:
[...]

PR #466, opened 2019-08-01
Allow GetFrame to suspend frame rendering. 

stuxcrystal:
Implements #324 

Some rules: Do not suspend and request frames at the same time. On error, use cancelSuspension to clean up.

Also: Visual Studio can suck dicks, so please make sure that I didn't overlook any tabs Visual Studio might have added.


src/core/vsapi.cpp
```diff
@@ -553,6 +553,26 @@ static void VS_CC getCoreInfo2(VSCore *core, VSCoreInfo *info) VS_NOEXCEPT {
     core->getCoreInfo2(*info);
 }
 
+static VSSuspensionContext* VS_CC suspendFrame(VSCore* core, VSFrameContext* frameCtx) VS_NOEXCEPT {
+    assert(core, frameCtx);
+	if (frameCtx->suspend) return 0;
```
FichteFoll: tab

src/core/vsapi.cpp
```diff
@@ -553,6 +553,26 @@ static void VS_CC getCoreInfo2(VSCore *core, VSCoreInfo *info) VS_NOEXCEPT {
...
+    assert(context);
+    context->core->threadPool->start(context->ctx);
+    delete context;
+}
+
+static void VS_CC cancelSuspend(VSSuspensionContext* context, VSFrameContext* frameCtx) VS_NOEXCEPT {
+	assert(context && frameCtx);
+	assert(context->ctx == frameCtx->ctx);
+	frameCtx->suspend = false;
+	delete context;
```
FichteFoll: tabs

src/core/vsthreadpool.cpp
```diff
@@ -216,10 +219,19 @@ void VSThreadPool::runTasks(VSThreadPool *owner, std::atomic<bool> &stop) {
 #ifdef VS_FRAME_REQ_DEBUG
             vsWarning("Exiting: %s Frame: %d Index: %d AR: %d Req: %d", mainContext->clip->name.c_str(), mainContext->n, mainContext->index, (int)ar, (int)mainContext->reqOrder);
 #endif
+			bool suspending = externalFrameCtx.suspend;
```
FichteFoll: tabs

HolyWu:
I would recommend users of Visual Studio simply changing the setting in **Tools -> Option -> Text Editor -> C/C++ -> Tabs** from **Keep tabs** to **Insert spaces** once and for all.

myrsloik:
I'll get to this one but it'll probably take a while.

sekrit-twc:
How does this feature work? Can you show an example sequence of API calls that a "suspended" filter would make?

stuxcrystal:
Of course.

Here is a completely useless example:
https://gist.github.com/stuxcrystal/a9867f17a44663330c1ae686c4a08c8d

sekrit-twc:
Why do we need a "VSSuspensionContext" type? Can't the frame context track if it 's suspended?

stuxcrystal:
Yes.

VSSuspensionContext is not used anywhere inside vapoursynth itself. It acts more like a token for the API-Consumer and is freed once the token is used (either by being cancelled or by resuming the frame). This way we can be sure that frame is only resumed once and cannot be used in subsequent suspensions of the same frame.

sekrit-twc:
Is there a limit to the number of suspended operations? It looks like filters may have to be aware of this to avoid overwhelming the underlying resource.

stuxcrystal:
There is no limit of suspended operations. Filters are free to use queues and semaphores for avoiding the underlying resource.

Renderers like vspipe and `VideoNode.output` only request a limited number of frames concurrently as they otherwise would request all frames at once thereby completly filling ram (This scenario would already happen, just try to request all frames at once using successive `VideoNode.get_frame_async`-calls for all frames on python.)

stuxcrystal:
I don't want to rush but do we get an ETA for your decision/input, myrsloik?

myrsloik:
Basically I'm hoping to get a first draft of audio support in before this and see where that goes. Unfortunately people keep giving me money to actually do other things...

myrsloik:
One more reason I remembered: It'd be nice to implement cancellation of frame requests so vapoursynth can abort properly if the underlying resource gets stuck. So that probably has to go in first as well.

stuxcrystal:
I kinda see no direct connection between audio and this.

The latter one is understandable though. Are you working on cancelation right now?

myrsloik:
No, audio first. Actually I spend every minute working on paid stuff now because of deadlines.

myrsloik:
Too many changes to merge this now. Will lump it into some general improvement issues instead.

PR #482, opened 2019-08-11
SSE2 optimization for misc.AverageFrames

sekrit-twc:
5% speedup on QTGMC.


couleurm:
Hey, sorry to dig out this old PR but, how is QTGMC and AverageFrames related?

PR #484, opened 2019-08-18
Expr

sekrit-twc:


src/core/exprfilter.cpp
```diff
@@ -18,540 +18,2481 @@
...
+
+        if (isOpCode(node, { ExprOpType::MUL, ExprOpType::DIV }))
+            return false;
+
+        // Deduce net sign of term.
+        const ExpressionTreeNode *parent = node.parent;
+        const ExpressionTreeNode *cur = &node;
+        int polarity = 1;
+
+        while (parent &
```
kblomster: Leftover debug code? Or is this some deep magic?

sekrit-twc: Thanks. Fixed.

stuxcrystal: If you think you resolved the issue.

Click on the "Resolve Conversation" button. Neither kblomster, nor anyone else afaik can do so.

If this turns out to be wrong, one can always start a new review. :)

PR #496, opened 2019-09-09
Basic audio tone support

dynamiccast:
This PR lays the foundation for audio support.

For this first step, we wanted to be able to hear a sound when playing a vpy on virtual dub on Windows.
VSVFW has been modify to declare an audio track
Vapoursynth has been modify to have a `GetAudio` function for each filter.
`Tone` is a function to produce a sounds of a certain frequency over a certain length. Output is always 16 bits.
`MixAudio` mixes to audio tracks
`FadeInOut` fades an audio track at the beginning and at the end
`AudioDub` adds an audio tracks to a video track

Here is a test script to try it out:

```
from vapoursynth import core

c=C=261.6255653006
D=d=293.6647679174
E=e=329.6275569129
F=f=349.2282314330
G=g=391.9954359817
A=a=440.00000000001
B=b=493.8833012561

def note(len, freq):
	return core.std.FadeInOut(core.std.Tone(length=len,frequency=freq,channels=2,level=1.0,samplerate=48000), 500)


video = core.std.BlankClip(length=42)
audio = note(0.25, A) + note(0.25, B) + note(0.25, C) + note(0.25, D) + note(0.25, E) + note(0.25, F) + note(0.25, G)

track = core.std.AudioDub(video, audio);

track.set_output()
```

This is obviously still a work in progress and may be used by someone to build a more robust audio support for VapourSynth.

### DIFF
--- a/include/VapourSynth.h
+++ b/include/VapourSynth.h
@@ -137,7 +137,8 @@ typedef enum VSPresetFormat {
     /* special for compatibility, if you implement these in any filter I'll personally kill you */
     /* I'll also change their ids around to break your stuff regularly */
     pfCompatBGR32 = cmCompat + 10,
-    pfCompatYUY2
+    pfCompatYUY2,
+    pfAudioOnly
 } VSPresetFormat;
 
 typedef enum VSFilterMode {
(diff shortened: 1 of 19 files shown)


DJATOM:
As I know, Myrsloik already working on audio support, branch `doodle1`.

myrsloik:
What @DJATOM said. Sitting there at half done due to other commitments.
Copying the one video and/or audio track model from avisynth is also just weird when you can redo things and have pure audio/video tracks and lump them together.
You also changed the API and ABI, thereby breaking all existing plugins. That itself ensures a rejection.

dynamiccast:
@myrsloik How do you suggest to tackle this without changing the API ? At some point, VSFW will have to call each filters to fill an audio buffer.

myrsloik:
See the doodle1 branch. You're free to add new functions to the API which won't break old things.

Issue #390, opened 2018-05-07
vsscript is not thread-safe.


stuxcrystal:
## What happens?
The vapoursynth-module tracks its ScriptEnvironment data by its ID.
If another filter tries to enter an environment these steps are executed:

1. It checks if vsscript is enabled. If not, skip all further steps.
2. It stores the current value of the `_environment_id`-variable onto the `_environment_id_stack`-list.
3. It sets `_environment_id` to the new environment id.

To exit the environment afterwards these steps are followed:
1. Store the last value stored on the `_environment_id_stack`-list to `_environment_id`

This is all well and good, until you consider that both variables are shared between _all_ threads. This is especially problematic as filters like `std.FrameEval` and `std.ModifyFrame` call back to Python code _from their respective worker thread_.

### Here is what might happen:
1. The editor opens a new script-environment and executes a script there.
2. The script uses a long running `std.FrameEval`-filter.
3. The script ends and registers an output where the above filter is inside the filter-tree.
4. The requests a frame triggering the FrameEval-callback.
5. The user restarts the script.
6. The editor opens a new script-environment and executes another script there.
7. The old FrameEval-callback is still running and now tries to allocate a new filter. It will fail as the previous script-environment reset the environment.

The same thing will almost certainly happen with editors that support previewing multiple files at once.

## How to reproduce?
I attached a zip-file with a proof of concept that causes `VideoNode.set_output` to attach itself to the wrong node (or worse, failing to create a filter). It consists of a module wrapping core vpy-functions (vsscript) and a script (confuse-environment.py) that reproduces the error.
[confuse-environment.zip](https://github.com/vapoursynth/vapoursynth/files/1980210/confuse-environment.zip)

## Why is this bad
VSScript is effectively unstable when more than one Environment is in use. This is traditionally the case in VapoursynthEditor. Especially since the core is managed by the Python Garbage Collector. If any module holds reference over the core, the core does not get freed.

## How to fix.
Make both variables thread-local.

myrsloik:
Interesting, I always assumed the GIL would make the operations kinda acceptable. Will review it tomorrow.

PR #556, opened 2020-04-11
Allow Script-Runners to control which environments are active and which not

stuxcrystal:
This PR allows Script-Hosts to implement their own management of VapourSynth-environments.

# Rationale
## Shortcomings of the current Thread-Local Approach
VSScript allows for multiple concurrent script environments to be alive at the same time. This lead to [concurrency problems](https://github.com/vapoursynth/vapoursynth/issues/390) as multiple environments fought over the which one is active. The solution was to move this value into a thread-local variable.

This introduced problems when vapoursynth users actually tried to use Multi-Threaded applications inside vsscript. In addition, a implementation-detail of vsscript was leaking into the code: The environment-id. Most of the vapoursynth wrapper does not need to know the actual environment id, they just need a pointer to the environment they are attached
to.

## Event-loops

Event-loops have existed in the Python-World [for a long time](https://twistedmatrix.com/trac/). With the introduction of [`asyncio`](https://docs.python.org/3/library/asyncio.html) the actual use of event-loops in python became wide-spread.

Until 3.7, Python lacked the ability to use Task-local-variables. The module [`contextvars`](https://docs.python.org/3/library/contextvars.html) implemented a library that allowed framework-authors to influence which context is active at any given callback. This is used in asyncio to implement Task-local variables.

VapourSynth is not aware of `contextvars` and thus will use thread-local storage.

## Custom Script-Runners

People who want to write their own script-runners (vsedit) usually better what concurrency requirements they actually have. There is no way for those to influence how the vapoursynth module manages which environment is actually active.

# Proposal

This proposal changes three things:
* This PR encapsulates the variables that are intrinsic to a script-environment into its own structure.
  Instead of the environment-id (which is an integer), it uses this object to control which environment the callbacks are attached to.
* It introduces a `EnvironmentPolicy`, a class that can be used to control how the vapoursynth module manages which is actually active any any given time.
* Changes to the `Environment` API that had some issues with undefined behaviour.

## Encapsulation of the Script-Environment state.

Right now, an environment of the module has two attribues: The currenct `Core` instance and the outputs attached to it. These were stored in four global variables. Two of those variables were only used if the VapourSynth module is running in standalone-mode and the other to were only used if vsscript was active. It depends on an exposed private variable `_using_vsscript`, which set of variables is actually used.

The change encapsulates these two variables into its own `EnvironmentData`-class and changes how the callback-functions refer to these variables.

## EnvironmentPolicy-Instances

A host-application can now implement an EnvironmentPolicy-class and make vapoursynth use it to manage which environment is active at any time. This gets passed special API-methods that allow it to create new environments at will.

Special considerations have been made so that the currently policy does not leak to the user. This prevents the user to access Policy-instances and mess with its internals.

### Lifecycle of an EnvironmentPolicy

```
Host-Application                            vapoursynth.pyx
       |                                           |
       | -------register_policy(my_policy)-------> |
       | <--my_policy.on_policy_registered(API)--- |
       |                                           |
       .                                           .
       .                                           .
       .                                           .
       |                                           |
       | <-----my_policy.on_policy_cleared()------ |
       |  (Called at-exit or when the policy is    |
       |   deactivated)                            |
       |                                           |
```

### A special API for EnvironmentPolicies

During the registration of a policy, a special object is passed to the policy, that allow it to create new environments at will and pass them to client code as well. This object is bound to the policy and all methods on it will fail once the policy is disabled.

The methods are as follows:

* `EnvironmentPolicyAPI.unregister_policy() -> NoReturn` unregisters the policy. Will cause all subsequent calls to this object to fail.
* `EnvironmentPolicyAPI.wrap_environment(EnvironmentData) -> Environment` wraps an environment into a wrapper object that can be passed
  to other parts of the application.
* `EnvironmentPolicyAPI.create_environment() -> EnvironmentData` creates a new environment. No further initialiation is neccessary.

### Entering an environment at will

VapourSynth will ask the policy every time it wants to request the current environment.
To do that, the `EnvironmentPolicy.get_current_environment() -> Optional[EnvironmentData]` is called. If this function returns `None`,
the module assumes no environment is active and will fail with a corresponding error.

VapourSynth has its own callbacks. These callbacks are attached to a specific environment that needs to be enabled while running
the callbacks. In order to do this, the module will call the function
`EnvironmentPolicy.set_environment(EnvironmentData) -> Optional[EnvironmentData]`.
This will activate the environment for the given callback and return the previously active environment-object.

```
Host-Application                            vapoursynth.pyx                                 User Script
       |                                           |                                             |
       |                                           | <---------Script does a FrameEval---------- |
       | <---my_policy.get_current_environment---- |                                             |
       | -------(The active environment X)-------> |                                             |
       |                                           |                                             |
       |                     (Create a Func object and pass it to the filter)                    |
       |                         [attach the EnvironmentData for X to it]                        |
       |                                           |                                             |
       |                                           | ------------FrameEval-Clip----------------->|
       .                                           .                                             .
       .                                           .                                             .
       .                       (Some time later, the new environment is Y)                       .
       .                                           .                                             .
       .                                           .                                             .
       |                                           |                                             |
       |                                           |                                             |
       |                                 Enter the environment X                                 |
       | <-----my_policy.set_environment(X)------- |                                             |
       | ------(The previous environment Y)------> | (Y is stored on the stack frame)            |
       |                                           |                                             |
       |                                           | -------------Run the callback-------------> |
       |                                           | <------------------------------------------ |
       |                                           |                                             |
       | <-----my_policy.set_environment(Y)------- |                                             |
       | ------(The previous environment X)------> | (X is ignored)                              |
       |                                 Exit the environment X                                  |
       |                                           |                                             |
```

### Changes to the Environment-API

Unfortunately, this proved to be incompatible to the old way this was done (Having its own stack inside a thread-local). A second API has been introduced to fix this problem. Users of the old API will still use a thread-local stack and a warning is printed to the console that undefined behaviour can occur in certain cases.

The undefined behaviour occurs when using the context-manager inside coroutines and generators (since the frame it runs in might be suspended and resumed in another thread/context) or if the functions __enter__ and __exit__ are called manually.

This PR does not introduce this problem!

Environment-objects now come in two variants:
* The old-style retrieved via `vpy_current_environment()`. Calling `vpy_current_environment()` will now emit a Python deprecation warning. These can be used directly as context managers, storing their state externally thus causing potential bugs due to unexpected codeflow inside the python script. `Environment.use()` however behaves just as expected.
* The new-style retrieved via `get_current_environment()`.
  Instances of new-style-environments must use the `Environment.use()`-context-manager.

### Predefined EnvrionmentPolicy-objects

VapourSynth comes with two EnvironmentPolicy instances:

* `StandaloneEnvironmentPolicy`, which only has one `EnvironmentData`-object. Any call to `set_environment` is ignored.
* `VSScriptEnvironmentPolicy`, which implements the `VPYScriptExport.id` to `EnvironmentData`-mapping as well everything that is needed so that vapoursynth still acts like it used to be.

### Examples

This example implements a `EnvironmentPolicy` that acts like the `StandaloneEnvironmentPolicy`. However this class has a `reset_core()`-function that disables the currently active environment and starts a new one.

```py
import vapoursynth as vs
from vapoursynth import core

class ReplacableEnvironmentPolicy(vs.EnvironmentPolicy):
    def __init__(self):
        self._current = None
        self._api = None
    def on_policy_registered(self, special_api):
        self._api = special_api
        self.reset_core()
    def on_policy_cleared(self):
        self._current = None
    def get_current_environment(self):
        return self._current
    def set_environment(self, env):
        if env is not self._current:
            raise RuntimeError("The chosen environment is dead.")
    def is_alive(self, env):
        return env is self._current

    def reset_core(self):
        # THIS IS A CUSTOM METHOD
        self._current = self._api.create_environment()

policy = ReplacableEnvironmentPolicy()
vs.register_policy(policy)


e1 = vs.get_current_environment()
print(e1 == vs.get_current_environment())

policy.reset_core()   # Enable a new environment globally.

print(e1 == vs.get_current_environment())
```

The output will be:
```
True
False
```

### DIFF
--- a/doc/pythonreference.rst
+++ b/doc/pythonreference.rst
@@ -565,7 +565,8 @@ Classes and Functions
    Once the method :method:`on_policy_registered` has been called, the policy is responsible for creating and managing environments.
 
    Special considerations have been made to ensure the functions of class cannot be abused. You cannot retrieve the current running policy youself.
-   The additional API-calls are only valid if the policy has been registered. Once the policy is unregistered, all additional API-calls will fail with a RuntimeError.
+   The additional API exposed by "on_policy_registered" is only valid if the policy has been registered.
+   Once the policy is unregistered, all calls to the additional API will fail with a RuntimeError.
 
    .. py:method:: on_policy_registered(special_api)
 
--- a/doc/pythonreference.rst
+++ b/doc/pythonreference.rst
@@ -570,9 +570,7 @@ Classes and Functions
    .. py:method:: use
 
       Returns a context-manager that enables the given environment in the block enclosed in the with-statement and restores the environment to the one
-      defined before.
-
-      The following code is safe in all cases.
+      defined before the with-block has been encountered.
 
       .. code::
       
--- a/doc/pythonreference.rst
+++ b/doc/pythonreference.rst
@@ -711,7 +711,7 @@ Classes and Functions
 
    Internal class that stores the context sensitive data that VapourSynth needs. It is an opaque object whose attributes you cannot access directly.
 
-   A normal user has now way of getting an instance of this object. You can only encounter EnvironmentData-objects if you work with EnvironmentPolicies.
+   A normal user has no way of getting an instance of this object. You can only encounter EnvironmentData-objects if you work with EnvironmentPolicies.
 
    This object is weak-referenciable meaning you can get a callback if the environment-data object is actually being freed (i.e. no other object holds an instance
    to the environment data.)
--- a/doc/pythonreference.rst
+++ b/doc/pythonreference.rst
@@ -523,8 +523,23 @@ Classes and Functions
 
         env = vpy_current_environment()
         # sometime later
-        with env:
+        with env.use():
           # Do stuff inside this env.
+
+   .. warning::
+
+      Environment-objects obtained using the :function:`vpy_current_environment` can directly be used as
+      as a context manager. This can cuase undefined behaviour when used in combination with generators and/or
+      coroutines.
+
+      This context-manager maintains a thread-local environment-stack that is used to restore the previous environment.
+      This can cause issues if the frame is suspended inside the block.
+      
+      .. code::
+
+         env = vpy_current_environment()
+         with env:
+              yield
 
    .. py:function:: is_single()
 
@@ -544,12 +559,46 @@ Classes and Functions
 
       Has the environment been destroyed by the underlying application?
 
+   .. py:method:: copy
+
+      Creates a copy of the environment-object. It will not copy the environment itself.
+
+      Added: R50
+
+   .. py:method:: use
+
+      Returns a context-manager that enables the given environment in the block enclosed in the with-statement and restores the environment to the one
+      defined before.
+
+      The following code is safe in all cases.
+
+      .. code::
+      
+         env = vpy_current_environment()
+         with env.use():
+             with env.use():
+                 pass
+
+      Added: R50
+
 .. py:function:: vpy_current_environment()
+
+   Deprecated. Use :function:`get_current_environment` instead.
 
    Returns an Environment-object representing the environment the script is currently running in. It will raise an error if we are currently not inside any
    script-environment while vsscript is being used.
 
    This function is intended for Python-based editors using vsscript.
+   This function has been deprecated as this function has undefined behaviour when used together with generators or coroutines.
+
+.. py:function:: get_current_environment()
+
+   Returns an Environment-object representing the environment the script is currently running in. It will raise an error if we are currently not inside any
+   script-environment while vsscript is being used.
+
+   This function is intended for Python-based editors using vsscript.
+
+   Added: R50
 
 .. py:class:: EnvironmentPolicy
 
@@ -567,6 +616,8 @@ Classes and Functions
    Special considerations have been made to ensure the functions of class cannot be abused. You cannot retrieve the current running policy youself.
    The additional API exposed by "on_policy_registered" is only valid if the policy has been registered.
    Once the policy is unregistered, all calls to the additional API will fail with a RuntimeError.
+
+   Added: R50
 
    .. py:method:: on_policy_registered(special_api)
 
@@ -586,17 +637,12 @@ Classes and Functions
 
       :returns: An :class:`EnvironmentData`-object representing the currently active environment in the current context.
 
-   .. py:method:: push_environment(env)
+   .. py:method:: set_environment(environment)
 
-      This method is called by the module to change the currently active environment. It should behave like a stack.
+      This method is called by the module to change the currently active environment.
 
-      :param id: The :class:`EnvironmentData` to enable in the current context.
-
-   .. py:method:: pop_environment()
-
-      This method is called by the module to remove the currently selected environment from the current context.
-
-      :returns: The :class:`EnvironmentData`-object that has just been popped or None if no ID was active.
+      :param environment: The :class:`EnvironmentData` to enable in the current context.
+      :returns: The environment that was enabled previously.
 
    .. py:method:: is_alive(environment)
 
@@ -606,6 +652,8 @@ Classes and Functions
 
    This class is intended to be used by custom Script-Runners and Editors. An instance of this class exposes an additional API.
    The methods are bound to a specific :class:`EnvironmentPolicy`-instance and will only work if the policy is currenty registered.
+
+   Added: R50
 
    .. py:method:: wrap_environment(environment)
 
@@ -630,6 +678,8 @@ Classes and Functions
    installed.
 
    If no policy is installed, the first environment-sensitive call will automatically register an internal policy.
+
+   Added: R50
    
    .. note::
 
@@ -642,11 +692,14 @@ Classes and Functions
       * Using :function:`clear_output`
       * Using :function:`clear_outputs`
       * Using :function:`vpy_current_environment`
+      * Using :function:`get_current_environment`
       * Accessing any attribute of :attribute:`core`
 
 .. py:function:: has_policy()
 
    This class is intended for subclassing by custom Script-Runners and Editors. This function checks if a :class:`EnvironmentPolicy` has been installed.
+
+   Added: R50
 
 .. py:attribute: _using_vsscript
 
@@ -662,6 +715,8 @@ Classes and Functions
 
    This object is weak-referenciable meaning you can get a callback if the environment-data object is actually being freed (i.e. no other object holds an instance
    to the environment data.)
+
+   Added: R50
 
 .. py:class:: Func
 
--- a/doc/pythonreference.rst
+++ b/doc/pythonreference.rst
@@ -553,16 +553,19 @@ Classes and Functions
 
 .. py:class:: EnvironmentPolicy
 
-   This class is intended for subclassing by custom Script-Runners and Editors. The instance controls which environment is in which context. Each environment must be associated
-   with a unique ID (integer) that identifies which environment is in use. The exact meaning of "context" is defined by the concrete EnvironmentPolicy.
+   This class is intended for subclassing by custom Script-Runners and Editors.
+   Normal users don't need this class. Most methods implemented here have corresponding APIs in other parts of this module.
+   
+   An instance of this class controls which environment is in which context.
+   The exact meaning of "context" is defined by the concrete EnvironmentPolicy. A environment is represented by a :class:`EnvironemtnData`-object.
 
    To use this class, first create a subclass and then use :function:`register_policy` to get VapourSynth to use your policy. This must happen before vapoursynth is first
    used. VapourSynth will automatically register an internal policy if it needs one. The subclass must be weak-referenciable!
+   
+   Once the method :method:`on_policy_registered` has been called, the policy is responsible for creating and managing environments.
 
    Special considerations have been made to ensure the functions of class cannot be abused. You cannot retrieve the current running policy youself.
    The additional API-calls are only valid if the policy has been registered. Once the policy is unregistered, all additional API-calls will fail with a RuntimeError.
-   
-   Normal users don't need this class. Most methods implemented here have corresponding APIs in other parts of this module.
 
    .. py:method:: on_policy_registered(special_api)
 
@@ -590,7 +593,7 @@ Classes and Functions
 
    .. py:method:: pop_environment()
 
-      This method is called by the module to remove the currently selected core from the current context.
+      This method is called by the module to remove the currently selected environment from the current context.
 
       :returns: The :class:`EnvironmentData`-object that has just been popped or None if no ID was active.
 
--- a/doc/pythonreference.rst
+++ b/doc/pythonreference.rst
@@ -534,6 +534,8 @@ Classes and Functions
 
       This context-manager maintains a thread-local environment-stack that is used to restore the previous environment.
       This can cause issues if the frame is suspended inside the block.
+
+      A similar problem also existed in previous VapourSynth versions!
       
       .. code::
 
--- a/doc/pythonreference.rst
+++ b/doc/pythonreference.rst
@@ -594,11 +594,11 @@ Classes and Functions
 
       :returns: The environment ID that has just been popped or None if no ID was active.
 
-   .. py:method: is_alive(id)
+   .. py:method:: is_alive(id)
 
       Is the environment with the given ID alive?
 
-   .. py:method: get_core(id)
+   .. py:method:: get_core(id)
 
       :param id: The environment-id to retrieve the core from.
       :returns: the :class:`Core`-object that is associated with the given ID. Return None if there is no core associated with the ID.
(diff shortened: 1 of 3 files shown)


myrsloik:
Very interesting improvement. But does it require Python 3.7 or later? That bit is unclear to me. I'll read through all of the code changes today.

stuxcrystal:
It does not. I did not want to bind VapourSynth to a specific context-management feature. Instead I created an API that allows third-party applications to implement their own context-management.

If an application wants to use `contextvars`, they need to implement their own `EnvironmentPolicy`. A corresponding `EnvironmentPolicy` using `ContextVars` could look like this (it is untested though):

```py
import vapoursynth
from contextvars import ContextVar, Token


active_environment = ContextVar(f'vapoursynth:active')


class ContextVarEnvironmentPolicy(vapoursynth.EnvironmentPolicy):
    def __init__(self):
        self._known = None
        self._api = None

    def create_environment(self) -> Tuple[vapoursynth.EnvironmentData, vapoursynth.Environment]:
        """
        Creates a new environment that can be activated at will.
        """
        env = self._api.create_environment()
        self._known.add(env)
        return env, self._api.wrap_environment(env)
    def destroy_environment(self, env: vapoursynth.EnvironmentData):
        """
        Disables the environment.
        """
        self._known.remove(env)

    def on_policy_registered(self, special_api):
        self._api = special_api
        self._known = set()
    def on_policy_cleared(self):
        self._api = None

    def is_alive(self):
        return env is in self._known
    def get_current_environment(self):
        return active_environment.get(None)
    def set_environment(self, env):
        if env is not self._known:
            raise RuntimeError("The chosen environment is dead.")
        token = active_environment.set(env)
        prev = token.old_value
        if token.old_value is Token.MISSING:
            prev = None
        return prev
```

sekrit-twc:
What is an environment? I tried reading the python documentation but it didn't clarify this question.

stuxcrystal:
You might have noticed in editors like VSEdit that `import`-statements behave like they are already imported if you rerun a script in it. But the `Core`-instance is different in each run and each of them has their own set of outputs.

I define everything that changes inside VapourSynth between different vsscript-runs as an "Environment".

sekrit-twc:
Shouldn't the solution be to create a new Python runtime each time a script is opened/closed? Does vsscript_EvaluteScript not do that? Perhaps that needs to be addressed.

stuxcrystal:
EvaluateScript does not. It just `exec`s the code in the main interpreter.

Creating multiple interpreters in the same process can be done for a long time, but it is a relarively unknown feature, it breaks a lot of stuff, Cython support for this has just been **reintroduced** in the current Cython version. And these "subinterpreters" as they are called are buggy.

Subinterpreters are actively worked on right now as they are actively considered as a serious contender on allowing multi-core CPU-bound cpython (but it feels like we have a long way to go).

This means, there simply was no reliable way to implement it like this, applications may now depend on thid behaviour and all that fun stuff.

stuxcrystal:
Well, I'm done with this now? You up for merging it?

myrsloik:
I will as soon as the regression fix only R50 is out.

Kamekameha:
This looks nice and all, but I believe this just overcomplicates the Python implementation.  After all, this stems from the fact that vsscript exists;  having two paths of execution and pretending in one that such thing as a script environment exists does not seem practical. This one is just putting a bandage on the that.

I don't think it's vapoursynth's responsibility to deal with how a script may be run. And in the end, a `VSCore` is all the environment it needs, so much so that the `VSAPI` even makes sure to pass one around every time it's needed.  If anything, making this core specifically a thread-local in Python across the entire module should make enough of an "environment". <sub><sub>(and even better would be to rework/replace vsscript :)</sub></sub>

Of course, all that said, I guess this PR in itself seems fine.  Not like I understand much of the possible use-cases of it or what exactly was broken before anyway.

myrsloik:
It fixes the callback problem which I've kinda been aware of for a long time and is extremely counter-intuitive since as @stuxcrystal says vsscript kinda promises separate environments and then doesn't deliver at all, including a nasty multithreaded issue. The policy itself is something most users won't have to muck around with themselves since a default matching the current one minus a few bugs is supplied, right?

stuxcrystal:
The current implementation is designed so that it matches the current behaviour of VSScript and Standalone.

Kamekameha:
Oh, what callback problem?

myrsloik:
Use vsscript to create script environment A and B. Evaluate a script in each of these that uses a callback that runs when requesting frames such as FrameEval and ModifyFrame.
Then use getFrameAsync to have frame requests queued both from environment A and B and watch as the problem happens. It's actually a fairly easy condition to trigger by accident if you write a simple GUI that can preview two different scripts at once.
There are a couple of variations on this but I think that's the fastest and most reliable way to trigger it.

Kamekameha:
But wasn't that fixed in a previous PR?

stuxcrystal:
Not every was fixed in a previous PR.

The PR #389 did actually fix a large part of the problem. However it doesn't cover the edge cases.

These edge cases mainly stem from the fact that A) _not all concurrent python code runs multi-threaded anymore_ and B) _that vapoursynth manages the currently active environment in a thread-local stack-object_ and C) to allow scripts to continue using multithreading I had to _expose an API to set the core in a specific thread_ (`vpy_current_environment() -> Environment`).

Essentially you have three groups of people here:
1. You are a normal script writer. You are not affected by any of this script-environment problem (mostly). You can ignore both #389 and this PR.
2. Your vpy-Script or Library uses multithreading. In this case, you have to use `vpy_current_environment()`. This PR changes the API a bit because of edge-cases in #389 that needed to be addressed in order to fix some edge-cases. This PR remains backward compatible but will print RuntimeWarnings and DeprecationWarnings once you get closer to the edge-cases.
3. You write an application that executes arbitrary third-party `.vpy`-Scripts (possibly more than one per process). These applications I call "script-runners". Right now, once you run more than one `vsscript`-Environment you almost immediately encounter the edge-cases.

Since groups 1 and 2 are not (or barely) affected, we will ignore them.

Let's look closer at the problems you encounter if you write a script-runner.

The general flow of such an application is the following:
```
      Main-Thread                                                            VapourSynth                            VS-Threadpool
(Doing important stuff)                                                   (Python Wrapper)                    (Inside the FrameServer)
           |                                                                      |                                        |
           |                                                                      |                                        |
           | ----- (Run the Script a.vpy) -----> Thread A                         |                                        |
           |                                        |                             |                                        |
           |                                (Run the script) <-- normal stuff --> |                                        |
           |                                        |                             |                                        |
           | <--------- Thread finishes ------------+                             |                                        |
           |                                                                      |                                        |
           | ---------(Get the outputs)-----------------------------------------> |                                        |
           | <========(List of outputs)========================================== |                                        |
           |                                                                      |                                        |
           | ---(Do some post-processing on the output)-------------------------> |                                        |
           | <==(Get a clip with the required post-processing)=================== |                                        |
           | ---(Get me Frame X async)------------------------------------------> |                                        |
           |                                                                      | ---------(Get Frame X Async)---------> |
           | <==(Future)========================================================= |                                        |
           |                                                                      |                                        |
           |                                                                      |                                        |
           |                                                                      | <--------(Here's Frame X)------------- |
           | <---(Future resolves)----------------------------------------------- |                                        |
           |                                                                      |                                        |
```

So, let's answer some fundamental questions about this diagram:
1. Why does the vpy-Script need to run in a separate thread from the main thread? The main-thread might run an event-loop potentially rendering the application unresponsive if the script runs for some time.
2. Why does the post-processing not run in a separate thread? The creation of additional filters is fast enough as not be noticable.
3. Why is the frame requested asynchronously? See no. 1 for the answer to this.

So, what happens if the application runs more than one `vpy`-Script at once in different `vsscript`-Environments?

Well, this now becomes a clusterfuck to handle. Modern event-loop-based programming in Python involves coroutines. Coroutines (and generators) suspend the Python stack at specific points (Everywhere where a `yield` or `await` is used). Event loops just resume these coroutines thus continuing the stack. 

Consider this example:
```py
from vapoursynth import *
import asyncio

a = get_env_1()
b = get_env_2()

async def f1():
    with a:
        print("f1", vpy_current_environment())
        await asyncio.sleep(1)
        print("f1", vpy_current_environment())

async def f2():
    await asyncio.sleep(0.5)
    with b:
        print("f2", vpy_current_environment())
        await asyncio.sleep(1)
        print("f2", vpy_current_environment())

async def main():
    await asyncio.gather(f1(), f2())

asyncio.get_event_loop().run_until_complete(main())
```

What is the expected result?

```
f1 <Environment 1>
f2 <Environment 2>
f1 <Environment 1>
f2 <Environment 2>
```

What is the actual result?
```
f1 <Environment 1>
f2 <Environment 2>
f1 <Environment 2>
f2 <Environment 1>
```

What happened?
To understand this, we need to see the actual codeflow. All this is happening in a single thread:
```
AsyncIO-EventLoop       F1       F2           VS-Wrapper
     |                                             |
     | ----------------> +                         |
     |                   | -----------(Set Env)--> |
     |                   |                         |    => f1 Env-Stack: [<Environment 1>]
     |                   | <---------------------- |
     | <-(sleep 1 sec)-- |                         |
     | -------------------------> +                |
     | <-(sleep 0.5sec)---------- |                |
     |                   |        |                |    (0.5sec pass)
     | -------------------------> |                |
     |                   |        | --(Set Env)--> |    
     |                   |        |                |    => f2 Env-Stack: [<Environment 1>, <Environment 2>]
     |                   |        | <------------- |
     | <-(sleep 1 sec)----------- |                |
     |                   |        |                |    (0.5sec pass)
     |-----------------> |        |                |
     |                   | -----------(Pop Env)--> |
     |                   |        |                |    => f1 Env-Stack: [<Environment 1>]
     |                   | <---------------------- |
     | <-(finish)------- +        |                |
     |                            |                |    (0.5sec pass)
     | -------------------------> |                |
     |                            | --(Pop Env)--> |
     |                            |                |    => f2 Env-Stack: []
     |                            | <------------- |
     | <-(finish)---------------- +                |
     |                                             |
```

This is a great problem for requesting frames asynchronously as the function `getFrameAsync()` receives a callback. VapourSynth-functions that take a callback always need to determine the current environment. In the above example, it is not guaranteed the "current" environment is always the correct one.

So when do these edge-cases occur? Well. If you use VapourSynth-module for anything other than to write filter-trees. And you run more than one environment at the same time. Those two things usually happens when you write a script-runner. You can work around this by making sure the stack frame is never suspended.

These script-runners are the target-audience for this PR as they already control which environment is active when. So let them give more control over it.

Kamekameha:
Now, what I fail to understand is when does it go from running each script environment in their own threads to running them in a single thread asynchronously.  Your example only shows the latter.

Isn't the whole point to run scripts in separate threads because the may block the event loop?  That's even what the _fundamental question_ no. 1 states.

> This is a great problem for requesting frames asynchronously as the function `getFrameAsync()` receives a callback. VapourSynth-functions that take a callback always need to determine the current environment. 

Ah, this reminds me—it'd be nice if the `VSAPI` provided of a way to query the core from a `VSNodeRef`.  Within the callback submitted to `getFrameAsync()` there's no simple way to retrieve the core that produced the call, other than including it with the `userData` that is.

But I guess I'm getting somewhat off-topic there.

stuxcrystal:
> This is a great problem for requesting frames asynchronously as the function getFrameAsync() receives a callback. VapourSynth-functions that take a callback always need to determine the current environment.

VapourSynth-functions that take a callback always need to determine the current environment for a simple reason:

The callback-function might call one of these functions:
* `vapoursynth.get_core() -> Core` (which is called by `vapoursynth.core.__getattr__`)
* `vapoursynth.get_output() -> Union[vapoursynth.VideoNode, vapoursynth.AlphaOutputTuple]`
* `vapoursynth.get_outputs() -> Mapping[str, Union[vapoursynth.VideoNode, vapoursynth.AlphaOutputTuple]]`
* `vapoursynth.clear_outputs() -> NoReturn`
* `vapoursynth.VideoNode.set_output() -> NoReturn`

All these query or modify the current environment. If the callback-function is guaranteed not to call these functions, the vapoursynth-callback might not need to store a reference to the current environment.

Kamekameha:
> All these query or modify the current environment.

Oh, I'm aware of that.  What I said at the end was more about the C API/`VSAPI`, which is why I also said I might be getting off-topic there.

stuxcrystal:
My explicit use-case is the following Problem:

Jupyter Notebook runs in an AsyncIO-event-loop, happily blocking during the execution of a notebook-cell.

I need to manually set the current environment if I wanted to provide a function to reset the core object.

And since I can be sure that the user A) _explicitely requested the new core_ and B) _only one environment is active at the same time_ killing multithreading performance is overkill.

myrsloik:
Doh, I merged things in the wrong order and created a conflict. Will merge it the second that's resolved.

myrsloik:
There appears to be several unresolved issues with this that renders avfs and vsvfw unusable. Reverting it returns things to mostly working. I can't figure out exactly where things go wrong.

PR #564, opened 2020-04-27
Add __version__ to the wrapper module.

stuxcrystal:
That allows the user to detect the desired vapoursynth version even if the DLL-version is mismatched.

### DIFF
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -44,6 +44,7 @@ try:
 except ImportError as e:
     typing = None
 
+__version__ = namedtuple("VapourSynthVersion", "release api_major api_minor")(50, VAPOURSYNTH_API_MAJOR, VAPOURSYNTH_API_MINOR)
 
 _using_vsscript = False
 cdef object _environment_state = ThreadLocal()


myrsloik:
It has to be manually bumped every time? There's no sneaky preprocessor way to get it automatically on build? I mean vapoursynth.h is included after all.

stuxcrystal:
`__version__` is a defacto standard in Python modules.
Usually a string, tuple or anything else. I use the release as a normal wrapper-user usually looks at the release number. Especially as the new releases might not include any new VapourSynth-APIs-Functions.

myrsloik:
Ok, makes sense. It'd still be ideal if it pulled in the vapoursynth version from version.h so I won't forget to change one more field.

Kamekameha:
I'd still name it `__api_version__` (or just `api_version`) as `__version__` should just match the one in the `setup.py` file.  It'd be a weird convention to couple the package version with libvapoursynth's API version in the same variable, they're bumped separately after all.

myrsloik:
Did my own version of it with the version coming from version.h and split version and api_version. See 8fa4f56.

Kamekameha:
The thing with https://github.com/vapoursynth/vapoursynth/commit/8fa4f569f6f5c8959597139a10508305a0a18135 is that it now depends on a private header to compile and will fail to do so from, e.g., source distributions made with `python setup.py sdist`.

If anything, I'd just set a separate version for the Python bindings as changes in the core are pretty much unrelated to the Python package itself.  Just `1.0.0` might be a good start.

Also, a function to query both libvapoursynth's API and release versions at runtime would be nice to have in the API.  Right now the former is not possible while the latter requires constructing a `VSCore` first.

myrsloik:
Right, doh. Guess I'll manually bump it then. I'd still like it to match the core version since officially I don't support mismatched versions even if technically it works most of the time

Kamekameha:
I see, fair enough.

PR #567, opened 2020-05-03
[Discussion wanted] Add an __all__-Attribute

stuxcrystal:
If you do `from vapoursynth import *`, the `__all__` attribute defines what gets actually included.

P.S. **Please don't merge it just yet.** Maybe others have additional ideas what should be included by when *-importing the python module.

### DIFF
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -44,6 +44,14 @@ try:
 except ImportError as e:
     typing = None
 
+__all__ = [
+  "core", "get_core",
+  "get_output", "get_outputs", "clear_output", "clear_outputs",
+  
+  "VideoFrame", "VideoNode",
+  "Format"
+]
+    
 __version__ = namedtuple("VapourSynthVersion", "release_major release_minor")(50, 0)
 __api_version__ = namedtuple("VapourSynthAPIVersion", "api_major api_minor")(VAPOURSYNTH_API_MAJOR, VAPOURSYNTH_API_MINOR)
 
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -58,6 +58,7 @@ __all__ = [
     'YUV422P8', 'YUV422P9', 'YUV440P8', 'YUV444P10',
     'YUV444P12', 'YUV444P14', 'YUV444P16', 'YUV444P8',
     'YUV444P9', 'YUV444PH', 'YUV444PS', 
+  'NONE',
   'FLOAT', 'INTEGER',
   
   'get_output', 'get_outputs',
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -45,11 +45,25 @@ except ImportError as e:
     typing = None
 
 __all__ = [
-  "core", "get_core",
-  "get_output", "get_outputs", "clear_output", "clear_outputs",
+  'COMPAT',
+    'COMPATBGR32', 'COMPATYUY2',
+  'GRAY',
+    'GRAY16', 'GRAY8', 'GRAYH', 'GRAYS', 
+  'RGB',
+    'RGB24', 'RGB27', 'RGB30', 'RGB48', 'RGBH', 'RGBS',
+  'YCOCG', 'YUV',
+    'YUV410P8', 'YUV411P8', 'YUV420P10', 'YUV420P12',
+    'YUV420P14', 'YUV420P16', 'YUV420P8', 'YUV420P9',
+    'YUV422P10', 'YUV422P12', 'YUV422P14', 'YUV422P16',
+    'YUV422P8', 'YUV422P9', 'YUV440P8', 'YUV444P10',
+    'YUV444P12', 'YUV444P14', 'YUV444P16', 'YUV444P8',
+    'YUV444P9', 'YUV444PH', 'YUV444PS', 
+  'FLOAT', 'INTEGER',
   
-  "VideoFrame", "VideoNode",
-  "Format"
+  'get_output', 'get_outputs',
+  'clear_output', 'clear_outputs',
+  
+  'core', 
 ]
     
 __version__ = namedtuple("VapourSynthVersion", "release_major release_minor")(50, 0)


DJATOM:
Will it import all YUV420Px stuff or break it? I'm usually explicitly import needed stuff into scripts, but if your change will break it, that's actually bad.

stuxcrystal:
If you do `from vapoursynth import YUV420P8` for example, this will still work (and is in fact not influenced by this PR)

I marked the PR as a draft because it might break this:
```py
from vapoursynth import *
print(YUV420P8)
```

sekrit-twc:
What is special about the chosen keywords? Why not just leave it as-is?

OrangeChannel:
I don’t think we even need an `__all__` attr here, as the only real reason for it is to prevent cluttering one's namespace with non-private objects when doing a wildcard import. Seeing as the `vapoursynth` * namespace is pretty large `len([i for i in dir(vapoursynth) if i[0] != '_']) = 85`, and that wildcard imports should be *discouraged* for modules, who will even be using this `__all__`? I would also argue that the classes like VideoNode etc. shouldn’t be included as for a script-writer using the wildcard import, these are useless and would only be used in conditionals or in type hinting (who type hints in a vsscript?). 

The `__all__` in VS would have to be very long for it to be useful in cleaning up a wildcard namespace of non-private things vapoursynth defines that a script user probably shouldn’t/wouldn't be using directly, specifically: `['AlphaOutputTuple', 'CallbackData', 'ColorFamily', 'Core', 'Environment', 'Error', 'Format', 'Fraction', 'FramePtr', 'Func', 'FuncData', 'Function', 'Iterable', 'Mapping', 'MappingProxyType', 'NONE', 'Plugin', 'PresetFormat', 'RawCallbackData', 'SampleType', 'ThreadLocal', 'VideoFrame', 'VideoNode', 'VideoPlane', 'VideoProps', 'construct_signature', 'ctypes', 'gc', 'get_core', 'inspect', 'namedtuple', 'os', 'set_message_handler', 'sys', 'threading', 'traceback', 'typing', 'vpy_current_environment']`. This leaves us with an `__all__` thats 47 elements long.

stuxcrystal:
> Do we even want people to import * from vapoursynth?
Well you can't stop them, can you?

myrsloik:
I'm learning more python every day here with these pull requests but it sure does sound like a horrible workaround to prevent users from killing themselves.

stuxcrystal:
As I just said, `vapoursynth` is actually a prime target for star-importing.

The target audience of this module so to speak are script-writers who depend solely on this module (and some third-party helpers for this module). They seldomly use other (non-vapoursynth-related) modules. Cluttering the script-namespace is therefore a non-issue as nobody should ever depend on another script.

I do think however, we should define what gets imported into the script-namespace when star-importing vapoursynth.

First, as it essentially gives a list of "essential" functions, constants and classes to the reader of the module itself since vapoursynth-scripts only need a small subset of exposed module functions (as @OrangeChannel said, the format-constants and `core`). 

Second, it discourages them from directly using certain features that are not directly intended for script-writers by not importing them (`vpy_current_environment` comes to mind for example).

In the end I think it would be nice for them to just `from vapoursynth import *` and not need that clutter caused by having to reference the vapoursynth module for every call to core or for format-conversions.

stuxcrystal:
On a separate note (which is why I make a double post):

Do we actually need `get_output`, `get_outputs`, `clear_output` and `clear_outputs` in `__all__`?

OrangeChannel:
Probably not, but I wasn't 100% sure if those might be useful for some editor/previewer for vs scripts.

myrsloik:
> Do we actually need `get_output`, `get_outputs`, `clear_output` and `clear_outputs` in `__all__`?

I think they should be included since they're related to set_output().

Kamekameha:
> Do we actually need `get_output`, `get_outputs`, `clear_output` and `clear_outputs` in `__all__`?

I don't think we need them in, either.  Those are exactly useful only within script environments clients and, in my opinion, as soon as we could rid of that interface the better.

> I'm learning more python every day here with these pull requests but it sure does sound like a horrible workaround to prevent users from killing themselves.

Agree.  Explicit is better than implicit after all.

In all, a `__all__` list should limit itself to usable API and given that most of the `vapoursynth` module's types and the bindings in general are not **immediately** useful then such list is just extraneous info and could easily convey the wrong idea of what's really usable public API.

sekrit-twc:
Pretty sure the only "API" that users frequently interact directly with is... `core`

OrangeChannel:
> In all, a `__all__` list should limit itself to usable API and given that most of the `vapoursynth` module's types and the bindings in general are not **immediately** useful then such list is just extraneous info and could easily convey the wrong idea of what's really usable public API.

I think stux's reasoning here was there are two very different end users of the module, throwaway encoding script writers, and function writers. This is *only* intended to sort of cleanup the need to have two ""import"" lines to initialize the core and constantly writing `vs.xxx` for **those writing encoding scripts**. The `__all__` has no effect on the "API" (?) that function writers use / will be using.

Kamekameha:
Yes, this won't break any scripts doing explicit imports, just star imports.

myrsloik:
> > Do we actually need `get_output`, `get_outputs`, `clear_output` and `clear_outputs` in `__all__`?
> 
> I don't think we need them in, either. Those are exactly useful only within script environments clients and, in my opinion, as soon as we could rid of that interface the better.

What would it be replaced with? You can't skip things simply because you dislike them a bit.

stuxcrystal:
Nothing. If the script actually needs it, the author is free to explicitely import it.

OrangeChannel:
Sorry, correction to https://github.com/vapoursynth/vapoursynth/pull/567#issuecomment-623165753, `'NONE'` should be included in the `__all__` as it's a PresetFormat defined here: https://github.com/vapoursynth/vapoursynth/blob/71df9aaf32e42f960bdff8e34d17cf44b462b621/src/cython/vapoursynth.pxd#L58
@stuxcrystal

stuxcrystal:
I just added `NONE` to the `__all__` list, even if I think that realistically it will only be used by *func-authors.

stuxcrystal:
As the discussion died, I marked this PR as ready.

PR #571, opened 2020-05-08
Actually deprecate get_core()

stuxcrystal:
It has been deprecated for 3 years now and people are still using it.
Let's get a little bit more aggressive.


myrsloik:
I agree but shouldn't we print an annoying warning message first for a version or two?

stuxcrystal:
That's actually the only thing this PR actually does.

FichteFoll:
https://docs.python.org/3/library/warnings.html
https://docs.python.org/3/library/exceptions.html#DeprecationWarning

PR #572, opened 2020-05-08
Make Format-class environment agnostic and make Preset-Formats act like actual Format-objects

stuxcrystal:
1. Environment-sensitive `Format`-instances are problematic:

IDs of Custom-Formats on VapourSynth are generated by incrementing a counter inside the VSCore-Class. This means, in Python, the ID of custom formats actually depend on the currently active environment. 

A function-library writer might be tempted to write:
```py
import vapoursynth as vs
from vapoursynth import core
RGB90 = core.register_format(vs.RGB, vs.INTEGER, 30, 0, 0)
```

This might become problematic in edge-cases as two Formats (of different cores) have different IDs. Unfortunately a `Format`-instance can't detect whether or not it is used in the correct format or not.

By dynamically registering the format as needed, the format instance can actually be shared across environments.

2. `PresetFormat` instances now behave like `Format` instances. This makes it easier to use. We could write:

```py
import vapoursynth as vs
from vapoursynth import core
RGB90 = vs.RGB24.replace(bits_per_sample=30)
```

3. I defaulted `subsampling_w` and `subsampling_h` to `0` in `Core.register_format`. This allows us to write:

```py
import vapoursynth as vs
from vapoursynth import core
RGB90 = core.register_format(vs.RGB, vs.INTEGER, 30)
```


myrsloik:
Nothing is guaranteed once you start mixing environments and things obtained from them. This kinda makes it look like a legitimate thing to do. Passing frames, clips, functions and so on between them will also end very badly...

sekrit-twc:
This should not be implemented in the Python wrapper. If this "feature" is to be implemented, it should be done within VSCore.

FichteFoll:
I agree that this should probably be handled in the internal API instead of in a wrapper, but could someone briefly explain to me why formats are required to be registered and mapped to an id in the first place? Why would using a struct containing the actual parameters not work? Or a custom bitmask, which is basically the same but in an integer if this is a requirement? I've just never understood why this was done and nobody could answer this to be so far.

stuxcrystal:
> Nothing is guaranteed once you start mixing environments and things obtained from them. This kinda makes it look like a legitimate thing to do. Passing frames, clips, functions and so on between them will also end very badly...

When I first thought of this feature, I actually thought, adding an API so it does not look like the format comes from the core while at the same time deprecating the existing `Core.get_format` and `Core.register_format`-function.

These two APIs I had in mind:
```py
import vapoursynth as vs

# Idea 1:
RGB90 = vs.Format(vs.RGB, vs.INTEGER, 30)

# Idea 2:
RGB90 = vs.register_format(vs.RGB, vs.INTEGER, 30)
```

I ultimately decided against it because, as `get_core` has shown us, actually deprecating one of the core functions of VapourSynth will take a lot of time and might in the end prove futile. (First, many script writers never actually used python at all before, secondly, the same script writers tend to learn from tutorials and seldomly if ever RTFM and thirdly, there is considerable inertia in libraries that really getting rid of deprecated core functions might very well be impossible)

However in VapourSynth there is already precedent for Environment-agnostic formats: `PresetFormat`s. 
In the Python-wrapper, these preset formats behave radically different (PresetFormats are actually just `int`s) to actual format instances even though they represent the same thing.

The problem right now is, making PresetFormats behave more like actual format objects may lead other people to believe that you can just put your custom formats on top of your module. This is coupled with the `vs.core` suggesting the inexperienced script writer that there is always ever only one `Core`-object.

To reiterate:

1. The current `core`-Proxy does already suggest to inexperienced writers that the core is a real singleton. Which it isn't.

2. `PresetFormat`s and `Format`s represent the same thing but behave different. Why?

3. Making them behave more like each other will suggest library-writers to put their own custom formats at the top of their module.

My solution is to just fake being environment-agnostic.

---

> This should not be implemented in the Python wrapper. If this "feature" is to be implemented, it should be done within VSCore.

Reasonable. But I really don't feel comfortable writing C or C++. So I did it in Python.

My solution does not

---

> Or a custom bitmask, which is basically the same but in an integer if this is a requirement?

Let me quote from [#encode-autism](https://discordapp.com/channels/221919789017202688/224104675031384065/708361220079616000) on [the Weeb-Autism Discord](https://discord.gg/ZB7ZXbN):
```
[18:54] Frechdachs: I guess to make every part of a format defined by a bit mask, you'd need 21 bits
color family 8 bits (theoretical support for 256 formats)
[18:54] Frechdachs: 1 bits sample type (only float and int)
[18:56] Frechdachs: bitdepth also 8 (for up to 256 different bitdepths)
[18:57] Frechdachs: and 2 bits each for subsampling_w and subsampling_h
[18:57] Frechdachs: no actually you need at least 3 for subsampling
[18:58] Frechdachs: since valid value ranges are 0-4
[18:58] stux!: You're free to implement this by yourself
[18:58] Frechdachs: but tbh, incrementing the enum for the color family seems saner
[18:58] Frechdachs: as it is done now
[18:59] Frechdachs: because you'd have to make a lot of assumptions which might make adding things to formats in the future harder
[19:00] Frechdachs: >You're free to implement this by yourself
                    was just evaluating the possibilities, since some were wondering why VS does things differently
```

stuxcrystal:
> This should not be implemented in the Python wrapper. If this "feature" is to be implemented, it should be done within VSCore.

Why not keep this list globally? Two formats from two different cores don't really differ at all. They, after all, still hold the same data.

sekrit-twc:
>But could someone briefly explain to me why formats are required to be registered and mapped to an id in the first place?

It is a good question. I think format IDs should just be deprecated. The only place where the API needs to receive a format specification from the caller, `newVideoFrame` already takes a `VSFormat *`.

    // Deprecated. Replaced with...
    const VSFormat *(VS_CC *registerFormat)(int colorFamily, int sampleType, int bitsPerSample, int subSamplingW, int subSamplingH, VSCore *core) VS_NOEXCEPT;

    // One of the use cases for registerFormat is to validate that the format is legal.
    int (VS_CC *isValidFormat)const VSFormat *format) VS_NOEXCEPT;

Frechdachs:
Since I got quoted I might aswell post too...

I do agree with sekrit-twc and FichteFoll, that if something like that is implemented, it should be done in internally and not in the python bindings.

And for the sake of completeness, this is a simple example where the edge case stux mentioned happens, when the user would not expect that he is mixing environments:

example module bad_module.py
```python
import vapoursynth as vs
YUV422S = vs.core.register_format(vs.YUV, vs.FLOAT, 32, 1, 0)

# some functions that use this format
```

example script
```python
import vapoursynth as vs
import bad_module as bad

YUV420S = vs.core.register_format(vs.YUV, vs.FLOAT, 32, 1, 1)
YUV422S = vs.core.register_format(vs.YUV, vs.FLOAT, 32, 1, 0)

raise Exception(f"module: {bad.YUV422S.id}, script: {YUV422S.id}")
```
In VSEdit the first press of F5 will print the same value, but every subsequent press will output different IDs for the same format.

(But yes, the name `register_format` might imply that it is not a good idea to use it at the top of a python module.)

sekrit-twc:
Thinking some more about backwards compatibility, there is code out there that relies on comparing `VSFormat::id` to check if two formats are the same. In light of this, to decouple VSFormat from VSCore, there needs to be a deterministic serialization of `VSFormat`, which is what @stuxcrystal proposed.

    // Serialization of VSFormat for backwards compatibility. 13 bits used.
    struct VSFormatId {
        compatFormat : 1; // Set if cmCompat
        colorFamily : 2; // 0 = cmGray, 1 = cmRGB, 2 = cmYUV, 3 = cmYCgCo. If compatFormat=1, 0 = pfCompatBGR32, 1 = pfCompatYUY2
        float : 1; // 0 = stInteger, 1 = stFloat
        log2BytesPerPixel : 2; // 0 = byte, 1 = word, 2 = dword, 3 = illegal
        paddingBits : 3; // (1<< log2BytesPerPixel) * 8- bitDepth. 0 for floats
        log2subsamplew : 2;
        log2subsampleh = 2;
    };

For the Python wrapper, the YUV420P8, etc. variables should be replaced with actual VSFormat objects. The vpy wrapper for `core.register_format` can be stubbed to construct format objects.

In VSAPI, `registerFormat` needs to be retained for backwards compatibility, but a new `createFormat` function should be added that does not modify the `VSCore` state. It takes care of calculating the serialized format and setting the format name. The Python wrapper can delegate to this when initializing the new YUV420P8 variables.

    // Deprecated.
    const VSFormat *(VS_CC *registerFormat)(int colorFamily, int sampleType, int bitsPerSample, int subSamplingW, int subSamplingH, VSCore *core) VS_NOEXCEPT;

    // New API. Returns 1 if invalid, 0 if valid.
    int (VS_CC *createFormat)(int colorFamily, int sampleType, int bitsPerSample, int subSamplingW, int subSamplingH) VS_NOEXCEPT; // No VSCore parameter! Stateless!

myrsloik:
I decided that VSFormats are singletons so you can do pointer comparison to check for format equality. The the predefined id bits are mostly to have some easier way to get common formats and to not have to pass around an actual pointer address as the "unique id".

We're stuck with the current system anyway or the ABI of everything will break.

myrsloik:
> Thinking some more about backwards compatibility, there is code out there that relies on comparing `VSFormat::id` to check if two formats are the same. In light of this, to decouple VSFormat from VSCore, there needs to be a deterministic serialization of `VSFormat`, which is what @stuxcrystal proposed.
> 
> ```
> // Serialization of VSFormat for backwards compatibility. 13 bits used.
> struct VSFormatId {
>     compatFormat : 1; // Set if cmCompat
>     colorFamily : 2; // 0 = cmGray, 1 = cmRGB, 2 = cmYUV, 3 = cmYCgCo. If compatFormat=1, 0 = pfCompatBGR32, 1 = pfCompatYUY2
>     float : 1; // 0 = stInteger, 1 = stFloat
>     log2BytesPerPixel : 2; // 0 = byte, 1 = word, 2 = dword, 3 = illegal
>     paddingBits : 3; // (1 << log2BytesPerPixel) * 8 - bitDepth. 0 for floats
>     log2subsamplew : 2;
>     log2subsampleh : 2;
> };
> ```
> 
> For the Python wrapper, the YUV420P8, etc. variables should be replaced with actual VSFormat objects. The vpy wrapper for `core.register_format` can be stubbed to construct format objects.
> 
> In VSAPI, `registerFormat` needs to be retained for backwards compatibility, but a new `createFormat` function should be added that does not modify the `VSCore` state. It takes care of calculating the serialized format and setting the format name. The Python wrapper can delegate to this when initializing the new YUV420P8 variables.
> 
> ```
> // Deprecated.
> const VSFormat *(VS_CC *registerFormat)(int colorFamily, int sampleType, int bitsPerSample, int subSamplingW, int subSamplingH, VSCore *core) VS_NOEXCEPT;
> 
> // New API. Returns 1 if invalid, 0 if valid.
> int (VS_CC *createFormat)(int colorFamily, int sampleType, int bitsPerSample, int subSamplingW, int subSamplingH) VS_NOEXCEPT; // No VSCore parameter! Stateless!
> ```

I like this idea. A sane packed version in uint32_t (so a few bits are unused/reserved if I ever want to extend things. This shall be put in the general API issues and I intend to fix it in the audio/experimental stuff in general branch.

stuxcrystal:
Until https://github.com/vapoursynth/vapoursynth/pull/572#issuecomment-634212935 is implemented, we can provide a similar functionality in the vapoursynth module only.

That being said, I changed the proposed API in a way that does not rely on the current core:

```py
>>> import vapoursynth as vs

# Creating new Format-objects is easier now. Subsampling-values are defaulted to 0
>>> vs.Format(vs.RGB, vs.FLOAT, 32)
<Format family=RGB sample_type=Float bits_per_sample=32 subsampling_w=0 subsampling_h=0>
>>> isinstance(vs.RGB24, vs.Format)
True
>>> vs.Format(vs.RGB, vs.INTEGER, 8) == vs.RGB24
True

# You can still pass subsampling if desired.
>>> vs.YUV420P8
<PresetFormat 'YUV420P8': 3000010>
>>> vs.YUV420P8.format
<Format family=YUV sample_type=Integer bits_per_sample=8 subsampling_w=1 subsampling_h=1>
>>> vs.Format(vs.YUV, vs.INTEGER, 8, 1, 1)
<Format family=YUV sample_type=Integer bits_per_sample=8 subsampling_w=1 subsampling_h=1>

# PresetFormats act like formats.
>>> vs.RGB24
<PresetFormat 'RGB24': 2000010>
>>> vs.RGB24.bits_per_sample
8

# You can still pass subsampling if desired.
>>> vs.YUV420P8
<PresetFormat 'YUV420P8': 3000010>
>>> vs.YUV420P8.format
<Format family=YUV sample_type=Integer bits_per_sample=8 subsampling_w=1 subsampling_h=1>
>>> vs.Format(vs.YUV, vs.INTEGER, 8, 1, 1)
<Format family=YUV sample_type=Integer bits_per_sample=8 subsampling_w=1 subsampling_h=1>

# You can now use replace(...) on PresetFormats
>>> vs.RGB24.replace(bits_per_sample=9)
<Format family=RGB sample_type=Integer bits_per_sample=9 subsampling_w=0 subsampling_h=0>
```

This makes it easier to create new formats. It automatically supports multiple-environments by automatically passing the correct ID depending on the currently activated environment.

Using this new syntax reduces bugs by new library writers unaware of the issues posed by multiple Environments. The environment-specific ID is only returned if explicitly requested by `int(format_obj)` or `format_obj.id`. Both calls should be unnecessary in library-code.

stuxcrystal:
So any comments on my changed API? @myrsloik

myrsloik:
This pull request is obsolete now, right? In doodle1 all format ids are a simple serialization of what they mean and subsampling now defaults to 0.

PR #581, opened 2020-05-31
Default to 0 if a python function returns None

stuxcrystal:


myrsloik:
Exactly what corner case does this fix?

stuxcrystal:
https://p.kageru.moe/j68LFw

`In[19]` to `In[21]`

PR #583, opened 2020-06-03
Automatically compile vapoursynth on push or PR.

stuxcrystal:
That way PRs that cause builds to fail won't be merged anymore.

P.S. I left the failing tests in. Should I remove the part that runs the tests in vapoursynth (since they haven't been updated in two years)?


myrsloik:
There are failing tests? I'll have to investigate that since they should all succeed

.github/workflows/c-cpp.yml
```diff
@@ -0,0 +1,47 @@
...
+      uses: actions/setup-python@v2
+      with:
+        # Version range or exact version of a Python version to use, using SemVer's version range syntax.
+        python-version: 3.8
+    - name: Install build dependencies
+      run: pip install cython
+    - name: configure
+      run: |
+        ./autogen.sh
+        LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$(command -v python | xargs dirname)/../lib LIBRARY_PATH=$LIBRARY_PATH:$(command -v python | xargs dirname)/../lib PKG_CONFIG_PATH=$(command -v python | xargs dirname)/../lib/pkgconfig ./configure --prefix=/usr/
```
FichteFoll, line 37: `$(command -v python | xargs dirname)/../lib` is used 3 times in this call and another 2 in the next step. Wouldn't it make sense to store this in a variable?

stuxcrystal:
@myrsloik https://github.com/stuxcrystal/vapoursynth/actions/runs/123568379
Look here for more info

.github/workflows/c-cpp.yml
stuxcrystal: Working on it

stuxcrystal:
> There are failing tests? I'll have to investigate that since they should all succeed
No problem. So I'll leave the ci-step in.

That means, I'm essentially done.

myrsloik:
There are two tests with python errors I'm not sure what to do about in prop_dict_test.py and weakref_test.py. Would you mind taking a quick look at them? That way this could get merged a little bit faster. I'll try to fix the core dump today.

myrsloik:
The core dump is fixed. Please take a look at the other python test failures now that this is merged.

stuxcrystal:
Working on it.

Issue #588, opened 2020-06-11
Environment policy appears broken


myrsloik:
The weakref test fails with an error that most likely indicates a problem with the new environment policy. A default environment isn't provided in some circumstances which makes weakref.py, vsvfw and avfs consistently fail with this error:

Python exception: cannot create weak reference to 'NoneType' object

Traceback (most recent call last):
  File "src\cython\vapoursynth.pyx", line 2517, in vapoursynth.vpy_evaluateScript
  File "src\cython\vapoursynth.pyx", line 2456, in vapoursynth._vsscript_use_environment
  File "src\cython\vapoursynth.pyx", line 355, in vapoursynth.use_environment
TypeError: cannot create weak reference to 'NoneType' object

stuxcrystal:
Can you check if my PR fixes the bugs. I actually never used avfs and vsvfsw.

**When does this error occur:**

If you pass `NULL` as ScriptEnvironment to  `vsscript_evaluate*`. It has its own environment initialization code different from `vsscript_createScript`. Because of this `vpy_createScript` was never called. That function was responsible to register the new instance to the VSScriptEnvironmentPolicy implementation. (Commit 1)

If you populate the ScriptEnvironment structure yourself and don't use `vpy_createScipt`. This is caused by the same problem. (Commit 2)

PR #589, opened 2020-06-11
Make sure vpy_createScript is always called.

stuxcrystal:
This is the most likely cause of #588


myrsloik:
I applied a version that fixes the recursive locking. Seems to fix things.

Issue #602, opened 2020-07-09
[Discussion] Arguments in VapourSynth APIv4


stuxcrystal:
I split up this discussion from #59 so other issues that could be addressed in APIv4 could be discussed over there.

This Post only holds the latest comment from myrsloik.

----

> While you're implementing APIv4:
> Perhaps it would be a good idea to somehow let plugins specify what they return in each function. This would solve problems with function-signature generation in Python.
> 
> For example, vsgenstubs currently just assumes every plugin-function just returns clips which is not the case.

Since we're on the subject of arguments. My plan at the moment is to have a way to enumerate all functions in a plugin (that vsmap string dump is horrible) and getting the plugin part of the info as well. Is returning the in and out arguments as a string good enough? Should that be improved too?

_Originally posted by @myrsloik in https://github.com/vapoursynth/vapoursynth/issues/59#issuecomment-656052783_

stuxcrystal:
I think this is good enough with strings.

I think we should provide the parameters in some form of a datastructure. That way, we could also add default values to the function signature.

myrsloik:
Not all things have static defaults. Actually about half the things have defaults that vary depending on other arguments and input clips. Not a fan of the added complexity with relatively limited benefits.

myrsloik:
See the last few commits in doodle1 for the somewhat improved api.

myrsloik:
How should V3 filters be handled (the case of no hint available)? Should I lie and say it's a single video node, report it as unknown or report it and "anything is possible"?

stuxcrystal:
I would say, anything is possible. Those who actually care about types will probably file appropriate issues to the plugin author, if issues can be filed against the project.

myrsloik:
Can you take a look at the argument bits in the doodle1 branch? It's nearing completion but I suspect I did mess up the return type hints since I have no idea what I'm doing. There are also some environment policy bits I'll create another issue for later...

PR #607, opened 2020-07-27
Actually use return_signature

stuxcrystal:
From request:
https://github.com/vapoursynth/vapoursynth/issues/602#issuecomment-664364461


myrsloik:
Looks ok I guess

myrsloik:
Did you take into consideration that the signature can be "any" for the return types? I don't see that change.

stuxcrystal:
_construct_type defaults `typing.Any` so it shouldn't pose a problem for unknown types

stuxcrystal:
I am however relatively sure return-types are not named.

PR #611, opened 2020-07-29
vapoursynth.pyx: Make get_core() return vs.core

thebombzen:
Superficially this should work fine, especially for simple vpy scripts, although I'm not entirely sure how it'll play with multiple environments


myrsloik:
It's deprecated. Use the "new" method instead. I don't see the point of this.

PR #612, opened 2020-07-30
vpy4 VSScriptOptions implementation

stuxcrystal:
Also includes my take on implementing logging.


myrsloik:
Looks good. Will try to compile it and test in a little bit.

stuxcrystal:
Note that there is another implementation for logging.

myrsloik:
I saw that. Having a separate logging path for vsscript is probably a must anyway. It also has less strict requirements (no need to keep a function reference alive and such).

PR #613, opened 2020-07-31
A faster concurrent VideoNode.frames-implementation (and using that to rewrite VideoNode.output)

stuxcrystal:
The implementation consists of two parts:
- Reimplementing `VideoNode.frames()` render multiple frames concurrently and yielding the frames in order.
- Reimplementing `VideoNode.output()` using the reimplemented `VideoNode.frames()`-method.

This makes `VideoNode.output()` easier to implement and its performance is now solely tied to `VideoNode.frames()` which has more uses than the `output`-function itself.

### DIFF
--- a/doc/pythonreference.rst
+++ b/doc/pythonreference.rst
@@ -356,16 +356,20 @@ Classes and Functions
       containing the *alpha* to output at the same time. Currently only vspipe
       takes *alpha* into consideration when outputting.
 
-   .. py:method:: output(fileobj[, y4m = False, prefetch = 0, progress_update = None])
+   .. py:method:: output(fileobj[, y4m = False, prefetch = 0, progress_update = None, backlog=-1])
  
       Write the whole clip to the specified file handle. It is possible to pipe to stdout by specifying *sys.stdout* as the file.
       YUV4MPEG2 headers will be added when *y4m* is true.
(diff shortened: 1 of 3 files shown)


stuxcrystal:
Well. I think I'm done.

stuxcrystal:
A few more bugs have been found.

luglio:
much appreciated

stuxcrystal:
ping

luglio:
Minor issue in case you will concern:
```python
>>> import vapoursynth as vs
>>> src = vs.core.std.BlankClip(width=10000, height=10000)
>>> for f in src.frames():
...     pass
... 
>>> exit()
Core freed but 300480000 bytes still allocated in framebuffers
```

myrsloik:
1. Memory leak?
2. Can you produce a version for the doodle1 branch instead?

stuxcrystal:
I can.

myrsloik:
Looks good. Final point to annoy you some more. I think AudioNode.frames() should get the same improvement as well. Don't even try to implement output() for it though since headers and packing is really annoying.

stuxcrystal:
You think so? https://docs.python.org/3/library/wave.html

myrsloik:
You still need to convert from "planar" audio to interleaved but I guess that could work.

stuxcrystal:
Merged with #619

PR #619, opened 2020-08-08
Improvments on logging and outputs

stuxcrystal:
Also implements #613 

Changes:
1. Implements #613 Including AudioNodes
2. StandaloneEnvironments (those that do not run inside vsscript or any other editor using EnvironmentPolicies), will their logging automatically bridged to the Python-`logging` facilities.
3. Inside VSScript-based applications, the Python `logging` facilities are redirected to the vapoursynth logging-facilities.
4. Inside VapourSynth, Python `warnings` are redirected to the vapoursynth logging-facilities.
5. Inside VSScript-based applications, `print` is redirected to the vapoursynth logging-facilities.
6. You can now manually override the `options`-dict in `EnvironmentData` with your own instance/subclass, allowing `EnvironmentPolicy`-subclasses to directly respond to `set_option()`-calls.
7. If you `VideoNode.output` to `sys.stdout` or `sys.stderr` in any `vsscript`/custom `EnvironmentPolicy`, it will error. Custom EnvironmentPolicies can work around that by wrapping the stream.

Clarification about points 2-5:
You have 3 cases:
1. You just want to use vapoursynth using vspipe, and vsedit, etc (or you are developing applications using vsscript in general):   VapourSynth will now automatically redirect stderr/stdout, warnings and logs from python's `logging`-module to your application to deal with.
2. You just want to import vapoursynth to do your own stuff with it and are using it from your own standalone python-script:  
VapourSynth will send its logs to the `logging`-module.
3. You just want to implement your own Python-based `EnvironmentPolicy`:  
You are on your own. VapourSynth will not assume how logging needs to be handled in your application.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -1,4 +1,4 @@
-name: Linux
+name: macOS
 
 on:
   push:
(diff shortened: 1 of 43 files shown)


myrsloik:
I don't think @cython.freelist() is the right thing to use. It does some object recycling but only AFTER they'd be freed by normal python. Sounds like a recipe for even more references hanging around.

stuxcrystal:
Freelists have been removed.

stuxcrystal:
I think that's all for now

Issue #519, opened 2019-12-05
[BUG] imwri.Write produces dark images


jkotra:
Image with `imwri.Write`

![imwri](https://user-images.githubusercontent.com/29590316/70250867-9465db00-17a4-11ea-8131-2aad8cf91bfc.png)

Script:

```python

import vapoursynth as vs

core = vs.core

clip = core.ffms2.Source(source="E:\Encoding WS\Movie/main.mkv",threads=6)

clip = core.resize.Spline36(clip, format=vs.RGB24, matrix_in_s='709',transfer_in_s='709',primaries_in_s='709',
                            prefer_props=True)

print(core.text.CoreInfo(clip))

frames = [0, 68477]


for f in frames:
    print(f)
    clip.get_frame(f)
    clip = core.imwri.Write(clip, "PNG64", "source_compare_%d.png",overwrite=True)

```
This is the image taken from `VapourSynth Editor` with default settings:

![709-editor](https://user-images.githubusercontent.com/29590316/70251059-e0188480-17a4-11ea-8630-7f4ed3558cfd.png)

default settings in VS editor:
![Capture](https://user-images.githubusercontent.com/29590316/70251097-edce0a00-17a4-11ea-8267-82f197977968.PNG)

why is this difference?

I want them to be same, essential reproduce VS editor image with script.

myrsloik:
What do you mean by default setting? you convert the image to rgb in the vs edit script, right?

jkotra:
> 
> 
> What do you mean by default setting? you convert the image to rgb in the vs edit script, right?

no, i did not convert anything.

my vsedit script:
```py
import vapoursynth as vs
core = vs.get_core()


clip = core.ffms2.Source(source="E:\Encoding WS\Movie/main.mkv",threads=6)
clip.set_output()
```


In VSedit preview window, settings can be accessed by clicking cogwheel button.

![tempsnip](https://user-images.githubusercontent.com/29590316/70263144-b028ab80-17bb-11ea-8a54-cd67fbbe0eb6.png)

myrsloik:
VSEdit will always invoke resize at the end to do a conversion to rgb with the default settings. 

`import vapoursynth as vs
core = vs.get_core()

clip = core.ffms2.Source(source="E:\Encoding WS\Movie/main.mkv",threads=6)
clip = core.resize.Spline36(clip, format=vs.RGB24, matrix_in_s='709',transfer_in_s='709',primaries_in_s='709',
[...]

PR #633, opened 2020-10-03
fix 'dark image' bug in imwri

jkotra:
This PR fixes bug i reported back in Dec 2019 (https://github.com/vapoursynth/vapoursynth/issues/519)

thanks to @motbob for finding out the cause of this bug. (https://github.com/vapoursynth/vapoursynth/issues/519#issuecomment-691735413)

## About bug

apparently, metadata corruption by `imagemagick`(`Magick++`) is causing this. this has nothing to do with vapoursynth itself, I tested it out using `Magick++` and it produced the same faulty output as VS.

## Solution

the solution is to strip the metadata before writing the file in ` vapoursynth/src/filters/imwri/imwri.cpp `.
```cpp
image.strip();
```

## Results

python script:
```py
import vapoursynth as vs

core = vs.core
core.std.LoadPlugin('/usr/local/lib/libffms2.so')
clip = core.ffms2.Source(source="video.mp4")

clip = core.resize.Spline36(clip, format=vs.RGB24, matrix_in_s='709',transfer_in_s='709',primaries_in_s='709',
                            prefer_props=True)

print(core.text.CoreInfo(clip))

frames = [0, 10, 20, 30]


for f in frames:
    print(f)
    clip.get_frame(f)
    clip = core.imwri.Write(clip, "PNG64", "source_compare_%d.png",overwrite=True)
```

Running the following will produce 3 images corresponding to given frames.

The example given here is **Frame 10** taken from [this](https://www.pexels.com/video/time-lapse-of-busy-street-857267/) video.

R52 (darker blacks compared to original):

This PR:

One way to compare is to open these images in separate tabs and switch between them. the difference should be obvious.

## Refs

* https://www.imagemagick.org/Magick++/Image++.html

### DIFF
--- a/src/filters/imwri/imwri.cpp
+++ b/src/filters/imwri/imwri.cpp
@@ -367,6 +367,8 @@ static const VSFrameRef *VS_CC writeGetFrame(int n, int activationReason, void *
                 writeImageHelper<uint8_t>(frame, alphaFrame, isGray, image, width, height, fi->bitsPerSample, vsapi);
             }
 
+            image.strip();
+
             image.write(filename);
 
             vsapi->freeFrame(alphaFrame);


myrsloik:
Interesting find. I looked at the source code and couldn't figure out what the "default" is if you don't strip the profiles. Will apply.

PR #637, opened 2020-10-23
Python 3.9 support for master branch

DJATOM:


DJATOM:
So updating PRs is not that difficult as I thought. Windows and Linux builds were successful, I assume everything else is OK.

myrsloik:
I think you missed updating vsscript.cpp

DJATOM:
In case you didn't received an update notification, I'm now posting that comment to help GitHub trigger such notification.

OrangeChannel:
Forgot to update the docs and the windows build instructions file on the readme @DJATOM

PR #638, opened 2020-10-23
Use Pynthon 3.9 for Windows workflow

DJATOM:


myrsloik:
But it won't produce a correct build. The python version is also hard coded in the source.

DJATOM:
I see. So I'll try to do that for previous PR then. I don't yet know how do update it tho, but maybe google help me.

PR #650, opened 2021-01-01
Fix undefined reference.

orbea:
The build fails with slibtool: (https://dev.midipix.org/cross/slibtool)
```
/usr/bin/ld: src/core/.libs/libvapoursynth_la-vsthreadpool.o: in function `VSThreadPool::spawnThread()':
vsthreadpool.cpp:(.text+0x2754): undefined reference to `pthread_create'
clang-11: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [Makefile:1259: libvapoursynth.la] Error 1
```
This is because vapoursynth sets `-no-undefined`, but GNU libtool will silently remove this allowing the build to succeed even with undefined references while slibtool actually respects `-no-undefined` correctly and fails to build.

This fixes ~~two related~~ one issue~~s~~.

1. It explicitly depends on `pthread` to fix the build.
2. ~~It uses `-Wl,-no-undefined` instead of `-no-undefined` so that even GNU libtool will respect it.~~

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -11,9 +11,11 @@ LT_INIT([win32-dll])
 AC_PROG_CC
 AC_PROG_CXX
 
+AC_CONFIG_MACRO_DIRS([m4])
 
 AC_SYS_LARGEFILE
 AC_FUNC_FSEEKO
+AX_PTHREAD
 
 
 AC_ARG_ENABLE([debug], AS_HELP_STRING([--enable-debug], [Enable compilation options required for debugging. (default=no)]))
--- a/Makefile.am
+++ b/Makefile.am
@@ -2,6 +2,7 @@ warningflags = -Wall -Wextra -Wno-unused-parameter -Wno-unused-function -Wno-ign
 commoncflags = -O3 -fvisibility=hidden $(warningflags) -I$(srcdir)/include $(DEBUGCFLAGS) $(MFLAGS) $(UNICODECFLAGS) $(STACKREALIGN)
 AM_CXXFLAGS = -std=c++11 $(commoncflags)
 AM_CFLAGS = -std=c99 $(commoncflags)
+ACLOCAL_AMFLAGS = -I m4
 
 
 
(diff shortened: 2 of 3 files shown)


orbea:
I removed the second commit changing `-no-undefined` to `-Wl,-no-undefined` because the OSX build fails.
```
ld: unknown option: -no-undefined
```
This does work with clang on linux though...

myrsloik:
Unclear if it fixes more than it breaks

orbea:
> Unclear if it fixes more than it breaks

This shouldn't break anything at all and is a real bug in vapoursynth...

orbea:
I attached logs of both the build with libtool and slibtool with this patch against the current git HEAD. It doesn't break anything on my end.

[vapoursynth.libtool.log](https://github.com/vapoursynth/vapoursynth/files/6372139/vapoursynth.libtool.log)
[vapoursynth.slibtool.log](https://github.com/vapoursynth/vapoursynth/files/6372142/vapoursynth.slibtool.log)

orbea:
Also for reference NetworkManager had the same issue which is now fixed for them.

https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/commit/3affccf29b53976a0cef5ea662f913d4913d44c8
https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/804

midipix:
Some background on why this fix _is_ absolutely necessary for as long as -no-undefined is kept:

* With no exception, all popular ELF-producing linkers (gnu ld, gnu gold, lld) permit unresolved symbols in the final image, which allows resolution of these symbols to remain the task/problem of the dynamic linker.

* Because this permissiveness leaves room for runtime problems, a strict linking mode, where all symbols must be resolved at link time, may be request via the linker flag --no-undefined.

* Unfortunately, by "swallowing" the above flag, gnu libtool gives users and developers the illusion that all symbols had been resolved, when in practice this often is not the case.

* My own view, as well as that of many in the community, is that gnu libtool's behavior in this particular respect is a serious bug which allows for build-logic bugs in projects such as vapoursynth to remain indefinitely hidden.

Ultimately, you'd need to decide whether ELF-specific assumptions may be hard-coded into vapoursynth's build logic. If the answer is yes, then simply remove -no-undefined and the illusion that comes with it (when building with gnu libtool). Otherwise, add -lpthread so that all symbols are indeed resolved at link time. The latter approach (adding -lpthread) is arguably the right, more portable, and more robust one, and so I'd recommend that you reconsider making the proposed change.

configure.ac
```diff
@@ -14,6 +14,7 @@ AC_PROG_CXX
 
 AC_SYS_LARGEFILE
 AC_FUNC_FSEEKO
+AX_PTHREAD
```
HolyWu, line 17: Consider adding `AC_CONFIG_MACRO_DIRS([m4])` to configure.ac too.

orbea: I am not sure if its preferable to use `AC_CONFIG_MACRO_DIRS([m4])` in `configure.ac`, `ACLOCAL_AMFLAGS = -I m4` in `Makefile.am` or both. Adding both is the only way to make `libtoolize` not complain, but all of the options seem to work fine.

I updated this PR to use both.

PR #657, opened 2021-01-27
Fix vspipe

DJATOM:


DJATOM:
Well, it pulled my syncing commit which was wrong, I'll make another PR

PR #667, opened 2021-03-19
doc: replace use of vs.get_core() with `from vapoursynth import core`

AkarinVS:
As the former is deprecated, it's better to recommend replacement in the docs.

Signed-off-by: akarin <AkarinVS@users.noreply.github.com>


theChaosCoder:
An alternative would be to change just `core = vs.core`

myrsloik:
Should be changed to vs.core I guess

AkarinVS:
PTAL.

PR #678, opened 2021-04-28
Pass undefined flag to the linker since libtool doesn't do it

HolyWu:


HolyWu:
Now the Linux CI actually fails, which is expected, due to undefined reference to `pthread_create` for reasons having been explained in https://github.com/vapoursynth/vapoursynth/pull/650.

PR #679, opened 2021-04-30
vspipe: honor sys.exit() exit code

### LINKED ISSUE #437: Featurerequest: sys.exit(code) does not work as intended in VSPipe
`sys.exit(code)` does not work as intended as it is essentially an alias for `raise SystemExit(code)`-Exception. All exceptions are being caught in vspipe and cause the exitcode 1 to be returned by vspipe instead of the exitcode provided by `sys.exit`.

While in other vsscript-applications catching SystemExit might be intended behaviour, vspipe only executes a single script and should therefore not replace an exitcode provided by the user.

Suggested change:
- Catch SystemExit in VSPipe and return the exitcode provided by its `code`-Attribute.

See: https://docs.python.org/3/library/exceptions.html#SystemExit

AkarinVS:
This PR modifies pyx to save the exit code to script's global
environment under key '__VS_exitCode'. This avoids exposing more
interfaces, but if deemed necessary, we could instead expose
the exception to VSScript interface as well.

Fixes #437.

Signed-off-by: akarin <i@akarin.info>

### DIFF
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -2208,6 +2208,7 @@ cdef public api int vpy_createScript(VPYScriptExport *se) nogil:
 cdef public api int vpy_evaluateScript(VPYScriptExport *se, const char *script, const char *scriptFilename, int flags) nogil:
     with gil:
         orig_path = None
+        exitCodeKey = '__VS_exitCode' # must match vspipe.cpp
         try:
             evaldict = {}
             if se.pyenvdict:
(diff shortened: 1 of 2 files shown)


myrsloik:
Interesting, I always thought a solution would need to be a bit more cluttered. I do however think amending the API will be cleaner. For example store the exit code in a new field in VPYScriptExport and then add an API function to get the script exit code.
Or actually since it's a behavior modification I'd rather do that in the doodle1 branch.

AkarinVS:
Do you think we also need a `isExitCodeValid` field?

The API I'm thinking looks like this:
Add this field to VPYScriptExport:
  int exitCode; // default to 0.

and then expose a new API:
  int vsscript_getExitCode(VSScript *handle)
which just returns the exitCode field.

What do you think?

Given that we're putting that to doodle1 branch, should I also
bump VSSCRIPT_API_MINOR?

Thanks.

AkarinVS:
Replaced with #680.

PR #680, opened 2021-05-02
vspipe: honor sys.exit() exit code (take 2 with new API)

### LINKED ISSUE #437: Featurerequest: sys.exit(code) does not work as intended in VSPipe

AkarinVS:
Introduced a new VSScript API, bumped minor version number.
(Probably unnecessary for VSScript4 as it's unreleased yet, but perhaps we do want
to move the new API to the last entry to not break existing software that is using the
current API?)

New API:
```C++
VS_API(int) vsscript_getExitCode(VSScript *handle);
```

Discussion: I also thought about adding a way to differentiate between regular failures
and `sys.exit(0)` but I don't think it's necessary, as it's not clear what the vpy author
intends when calling `sys.exit(0)`, vspipe will fail (and return 1) anyway.

Fixes #437.

### DIFF
--- a/doc/api/vsscript.h.rst
+++ b/doc/api/vsscript.h.rst
@@ -32,6 +32,8 @@ Functions_
 
    vsscript_getError_
 
+   vsscript_getExitCode_
+
    vsscript_getOutput_
 
    vsscript_clearOutput_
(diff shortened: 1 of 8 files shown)


myrsloik:
This isn't abandoned, just parked for a bit. The current plan is to only apply it to the doodle1 branch.

myrsloik:
I think the right thing to do is to only implement this in the doodle1 branch without bumping the api version.

AkarinVS:
Yes, this PR is against the doodle1 branch, and I have rebased and reverted
the change to the API version.

PTAL. Thanks.

PR #690, opened 2021-05-20
fix std.Expr memory leak

AkarinVS:
There are two issues:
1. base class ExprCompiler does not declare its destructor as virtual, therefore, when deleting through an ExprCompiler pointer, the destructor of ExprCompiler128/256 won't be called at all, thus leaking memory.
2. On Unix, `munmap(2)` requires the memory size to be non-zero (there is no need to aligned it to page size, though), otherwise it will silently fail with `EINVAL`, thus leaking code buffer memory.


src/core/exprfilter.cpp
```diff
@@ -231,7 +232,7 @@ class ExprCompiler {
     }
 
     virtual ~ExprCompiler() {}
-    virtual ExprData::ProcessLineProc getCode() = 0;
+    virtual ExprData::ProcessLineProc getCode(size_t &size) = 0;
```
sekrit-twc, line 235: Return a pair.

AkarinVS: Done. PTAL.

src/core/exprfilter.cpp
```diff
@@ -3327,7 +3330,7 @@ static void VS_CC exprCreate(const VSMap *in, VSMap *out, void *userData, VSCore
                             compiler->addInstruction(op);
                         }
 
-                        d->proc[i] = compiler->getCode();
+                        std::tie(d->proc[i], d->procSize[i]) = compiler->getCode();
```
sekrit-twc, line 3333: <tuple>

AkarinVS: what do you mean? could you please elaborate?

kblomster: @AkarinVS `std::tie()` is defined in `<tuple>` which is not included in this file, so I'd guess @sekrit-twc means you should add `#include <tuple>`. Don't take my word for it though.

AkarinVS: I see. All the checks have passed, so I never suspected a missing include. Done. PTAL.

PR #693, opened 2021-05-25
introduce sin to the expr filter

### LINKED ISSUE #541: Could you please add the sin () operation in Expr?
I want  make the solar curve, it used sin () operation

AkarinVS:
Fixes #541.

(shameless plug: for those interested, I still have a few other enhancements to `std.Expr` in https://github.com/AkarinVS/vapoursynth-plugin and I'd be more than happy to send PR to
to get some into `std.Expr`.)

### DIFF
--- a/doc/functions/expr.rst
+++ b/doc/functions/expr.rst
@@ -59,7 +59,7 @@ Expr
 
    The operators taking one argument are::
 
-      exp log sqrt abs not dup dupN
+      exp log sqrt sin cos abs not dup dupN
 
    The operators taking two arguments are::
 
@@ -84,6 +84,10 @@ Expr
       d < e
       
       abs(f)
+
+   The sin/cos operators are approximated to within 2e-6 absolute error for
+   inputs with magnitude up to 1e5, and there is no accuracy guarantees for
+   inputs whose magnitude is larger than 2e5.
 
    How to average the Y planes of 3 YUV clips and pass through the UV planes
    unchanged (assuming same format)::
--- a/src/core/exprfilter.cpp
+++ b/src/core/exprfilter.cpp
@@ -196,6 +196,8 @@ class ExprCompiler {
     virtual void exp(const ExprInstruction &insn) = 0;
     virtual void log(const ExprInstruction &insn) = 0;
     virtual void pow(const ExprInstruction &insn) = 0;
+    virtual void sin(const ExprInstruction &insn) = 0;
+    virtual void cos(const ExprInstruction &insn) = 0;
 public:
     void addInstruction(const ExprInstruction &insn)
     {
@@ -228,6 +230,8 @@ public:
         case ExprOpType::EXP: exp(insn); break;
         case ExprOpType::LOG: log(insn); break;
         case ExprOpType::POW: pow(insn); break;
+        case ExprOpType::SIN: sin(insn); break;
+        case ExprOpType::COS: cos(insn); break;
         default: vsFatal("illegal opcode"); break;
         }
     }
@@ -242,7 +246,7 @@ class ExprCompiler128 : public ExprCompiler, private jitasm::function<void, Expr
     friend struct jitasm::function_cdecl<void, ExprCompiler128, uint8_t *, const intptr_t *, intptr_t>;
 
 #define SPLAT(x) { (x), (x), (x), (x) }
-    static constexpr ExprUnion constData alignas(16)[39][4] = {
+    static constexpr ExprUnion constData alignas(16)[53][4] = {
         SPLAT(0x7FFFFFFF), // absmask
         SPLAT(0x80000000), // negmask
         SPLAT(0x7F), // x7F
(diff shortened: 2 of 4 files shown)


AkarinVS:
`cos` support is also available in my Expr fork. I can add it to this PR if deemed necessary to offer a symmetric API that includes both `sin` and `cos`.

Thanks.

myrsloik:
Interesting additions. I'll take a closer look later.

src/core/exprfilter.cpp
```diff
@@ -1647,7 +1667,46 @@ do { \
...
+            auto y = bytecodeRegs[insn.dst];
+            YmmReg t1, sign, t2, t3, t4;
+            // Remove sign
+            vmovaps(sign, ymmword_ptr[constants + ConstantIndex::absmask * 32]);
+            vmovaps(t1, sign);
+            vandnps(sign, sign, x);
+            vandps(t1, t1, x);
+            // Range reduction
+            vmovaps(t3, ymmword_ptr[constants + ConstantIndex::float_rintf * 32]);
+            vmulps(t2, t1, ymmword_ptr[constants + ConstantIndex::float_invpi * 32]);
```
sekrit-twc, line 1682: FMA?

doc/functions/expr.rst
```diff
@@ -85,6 +85,10 @@ Expr
       
       abs(f)
 
+   The sin operator is approximated to within 2e-6 accuracy for inputs with
```
sekrit-twc, line 88: relative or absolute error?

AkarinVS:
it's absolute error. I will revise the docs to make it clear.

AkarinVS:
using FMA requires another set of consts though. If that's OK, I will add a
FMA enabled version.

sekrit-twc:
Let's make the 256-bit version FMA-only. ExprFilter256 is already gated by AVX2.

AkarinVS:
converting `x cos` to `x pi/2 + sin` will work mathematically, but if x is much large than pi/2, then it will lose precision.

As you can see from my implementation here
https://github.com/AkarinVS/vapoursynth-plugin/blob/ff47ee4d86be4c4eb16f21f452c28a53fe47869e/expr/exprfilter.cpp#L1152
sin and cos are basically shared code, with different set of polynomial coefficients.

Therefore, I'd suggest that we add both `sin` and `cos` operators as a pair.
Please let me know your decision and I will make a revision incorporating all comments so far.

Thanks.

myrsloik:
Both sin and cos should be added and then I think I'll merge things

pinterf:
Nice additions, when I was porting Expr to Avs+ I tried to implement it, I found [_mm_sin_ps](https://software.intel.com/content/www/us/en/develop/documentation/cpp-compiler-developer-guide-and-reference/top/compiler-reference/intrinsics/intrinsics-for-short-vector-math-library-operations/intrinsics-for-trigonometric-operations/mm-sin-ps-mm256-sin-ps.html) and _mm_cos_ps but it was only available for Intel compilers (SVML library). I could not found its limitation on the input parameter and about the error magnitude of their implementation.

AkarinVS:
PTAL. I believe all comments have been addressed.
However, I'd like PR #706 to be merged first, so that we are sure that the
CI does run all the tests.

Thanks.

AkarinVS:
I should note that I have included three general bugfix and/or code clean
up commits for the expr filter at the end of this commit series.

commit b5862f5172b14a8375abd8f2adf4035216dc0487: fixes a missing case label
commit 3964e9d8093eb1f7ca8201e2251cf4884eb5bee4: fixes log(0) to return nan
instead of -87.3365, fixes #705.
commit 1a47c10e3ad66ef040403b6a0658425f0c61605a: removes two unnecessary
lines, probably copy-paste errors.

sekrit-twc (commented):
This commit isn't needed. There is no case where FMA is available but AVX isn't.

AkarinVS:
The commit is necessary, because there are CPUs with AVX but not FMA3 or
AVX2 (Intel Ivybridge),
and there are CPUs that have AVX&FMA3 but not AVX2 (AMD Piledriver).

sekrit-twc:
Those CPUs already can not run ExprCompiler256, because AVX2 is needed to run int-float conversions. Moreover, exp/log are already FMA-only.

AkarinVS:
Note the commit you're talking about is changing the implementation in the
ExprCompiler128 class, not ExprCompiler256, whose sin/cos implementation is
FMA only.

>

sekrit-twc:
I misread the commit. It still looks unnecessary, since the only CPU with FMA3 but not AVX2 is apparently PileDriver, which is both obsolete and useless (AVX/VEX incurs a huge performance penalty due to a CPU bug). I recommend leaving the commit out, since FMA is not used throughout the rest of Expr128.

AkarinVS:
please see
https://github.com/vapoursynth/vapoursynth/blob/7d6691db79bbe1fe8cd216640c59ae02e5561bfd/src/core/exprfilter.cpp#L583
.

sekrit-twc:
Yes, I stand corrected. Perhaps the fma() emitter should also be removed. I would rather not have two implementations of the transcendental function in essentially assembly, unless they can somehow be unified by macros.

sekrit-twc:
Also, in ExprCompiler128, please use a loop or subprocedure to execute sincos_(). Calling the emitter twice produces the bytecode twice (see exp/log), which is not desirable.

myrsloik:
So this is ready now?

AkarinVS:
Yes, the only thing is that I don't want to give up the SSE2 FMA3
optimization as I don't think the FMA and non-FMA code path is that hard to
maintain and I couldn't find an easy way to use macros to unify them.

sekrit-twc:
Leaving the code is fine, although I think a locally-defined macro can help for the FMA. Either way, the other comment should be addressed:

>Also, in ExprCompiler128, please use a loop or subprocedure to execute sincos_(). Calling the emitter twice produces the machine code twice (see exp/log), which is not desirable.

AkarinVS:
The latest commits are using a loop instead of expanding it two times for
ExprCompiler128.

PTAL.

sekrit-twc:
Looks fine. Thanks.

Issue #694, opened 2021-05-25
Installer creates invalid link to User Autoload Directory if directory didn't exist


luglio:
If the directory does not exist when VS is installed, it will be created but the link is broken. Otherwise it's fine.
![unknown](https://user-images.githubusercontent.com/13797847/119544997-e95e2e00-bd57-11eb-8ae7-1fcc1b8628b0.png)

myrsloik:
Installer limitation. Run vsrepo and it'll create the dirs.

PR #696, opened 2021-05-31
Manually create User Autoload Directory in Windows installer

luglio:
This should fix https://github.com/vapoursynth/vapoursynth/issues/694


myrsloik:
Actually this doesn't fix the problem and that's why it's not already  implemented this way. Counterexamples:

1. You have a computer with multiple users. This will only create the dir for the user who installs VS.
2. You are a normal user who provides admin account credentials to run the installer. The autoload dir will now be created for the admin user.

luglio:
I see. Creating dirs for every user might be easier without inno.

PR #717, opened 2021-07-27
CI: fix windows build

AkarinVS:
by removing `/D VS_USE_MIMALLOC`, `mimalloc-override.lib` and forced symbol reference `_mi_version` from visual studio project files.
The CI environment does not have mimalloc by default and generally mimalloc is not a documented dependency.

Of course, an alternative solution is to incorporate solution files for mimalloc, but that requires more extensive editing to the project files that could be done without actually using Visual Studio.

### DIFF
--- a/msvc_project/Core/Core.vcxproj
+++ b/msvc_project/Core/Core.vcxproj
@@ -87,18 +87,17 @@
       <WarningLevel>Level3</WarningLevel>
       <Optimization>Disabled</Optimization>
       <SDLCheck>true</SDLCheck>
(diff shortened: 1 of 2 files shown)


AkarinVS:
Unfortunately, this approach does not work...

myrsloik:
At this point I'm about to just give up with the free CI. In a period of many years it's at best worked well for 4 months or so. Once it's set up correctly it gets broken again by whatever third party service is used. It's at this point a complete waste of my time and I'll probably delete all traces of it soon.

AkarinVS:
I think we can keep Linux and macOS CI.
Those are much easier to maintain than the Windows CI.

myrsloik:
Remove the windows CI then as the solution. I do all development on windows anyway so it doesn't matter...

Issue #724, opened 2021-08-07
A frame's copy-on-write can invalidate Python references


Kamekameha:
https://github.com/vapoursynth/vapoursynth/issues/710#issuecomment-892268279

To illustrate:

```python
>>> a = ...  # a writable frame
>>> assert not a.readonly
>>> a0_1 = a.get_write_ptr(0)
>>> b = a.copy()  # shallow copy
>>> a0_2 = a.get_write_ptr(0)  # now COW triggers
>>> # the assertion here raises because of this even though
>>> # we requested for the same plane from the same frame
>>> assert a0_1.value == a0_2.value
>>> # this one also raises because a0_1 does in fact now live in frame reference b
>>> assert a0_1.value != b.get_write_ptr(0).value
>>> del b
>>> # after `del b` a0_1 is now a dangling pointer,
>>> # which *seems* unrelated considering we got a0_1 from `a` and `a` is still alive
```

Basically, pointers can move around frame references every time a copy might be needed.

This is true for every available method as of now to retrieve plane data from a frame, including `VideoPlane`.  Doubly so for `VideoPlane` since it doesn't even give the option to choose between `getReadPtr()` and `getWritePtr()`.  The buffer protocol relies on consistency on every buffer request, which this current behavior can't guarantee, and the main reason I want to deal with this before fixing/rewriting `VideoPlane`.

Because writability is handled on the Python end by `VideoFrame`, I suggest we trigger copy-on-write once specifically and assume ownership of the data afterwards.

Kamekameha:
So something like this:

```diff
diff --git a/src/cython/vapoursynth.pyx b/src/cython/vapoursynth.pyx
index 983705b1..6f1f8a02 100644
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -1209,7 +1209,16 @@ cdef class VideoFrame(object):
             self.funcs.freeFrame(self.constf)
 
     def copy(self):
-        return createVideoFrame(self.funcs.copyFrame(self.constf, self.core), self.funcs, self.core)
+        lib = self.funcs
+        frame = lib.copyFrame(self.constf, self.core)
+        # trigger copy-on-write to make sure we truly own the data, see GH-XXX
+        with nogil:
+            length = lib.getFrameFormat(frame).numPlanes
+            lib.getWritePtr(frame, 0)
+            if length == 3:
+                lib.getWritePtr(frame, 1)
[...]

Issue #423, opened 2018-11-25
Invalid output of VideoNode.output on exotic resolutions


stuxcrystal:
During testing of Yuuno 1.1, my testers reported a completely unrecognizable output when trying to encode clips with exotic resolutions.

Standard resolutions (720p, 1080p, etc.) work fine however.

Code for reproduction:

    c = core.lsmas.LWLibavSource(r"C:\Users\stuxc\Desktop\spyce.mkv")
    m = c.resize.Spline36(width=1516, height=854)[:500]
    with open('lol.y4m', 'wb') as f:
        m.output(f, y4m=True)

The video has then been encoded using the command `x264 --demuxer y4m lol.y4m -o test.mkv`. This is what happens

Windows 10 18.09 x64:
![applicationframehost_2018-11-25_19-39-25](https://user-images.githubusercontent.com/2260352/48983123-4b883a80-f0eb-11e8-8ae6-9319d4be9635.png)

Windows 10 18.03 x64 (different User)
![c7f7da](https://blubb-made.spdns.org/f/c7f7da.png)

Arch Linux:
![k53bc7](https://user-images.githubusercontent.com/2260352/48983128-69ee3600-f0eb-11e8-97e8-23758928aec4.png)

Kamekameha:
Yeah, I'm aware of it. This happens when width != stride (that is, when width is not mod 32) and so its block of memory is technically no contiguous; `io.RawIOBase.write` expects a contiguous buffer.

This was introduced by my `VideoPlane` thing not extensively checking what the flags passed to `VideoPlane.__getbuffer__()` are, this should raise an exception. I've been meaning to fix this since forever but given that nobody seems to use or report anything about these Python APIs I just kinda forgot about it.

myrsloik:
Sigh... you knew you added this bug when changing things around?

Kamekameha:
> At least I try to fix the bugs I make.

Good to know, I did say though that:

> I've been meaning to fix this since forever but given that nobody seems to use or report anything about these Python APIs I just kinda forgot about it.

> But will your fix make this raise an error or just make it a contiguous buffer?

I'll see.

stuxcrystal:
I'd like `VideoNode.output` to be able to output all resolutions. 

So if you decide against it, please revert this https://github.com/vapoursynth/vapoursynth/blob/d6c5e0317d2b3ab8e550773b09069b3904aa17e3/src/cython/vapoursynth.pyx#L494

[...]

PR #727, opened 2021-08-09
Refactor VideoFrame's buffer protocol support

### LINKED ISSUE #710: Cython memory views don't keep a reference to the underlying VSFrame
Exposed with the huge cython cleanup where frame references get freed faster.

Kamekameha:
A few things I think are worth mentioning:

About #423, I completely forgot about it and it's been ~3 years since I said I'd deal with that.
In any case, now planes are written line-by-line if the data is not contiguous.  This should be faster even with all the overhead of creating temporary objects.

I took the liberty to make VideoFrame a sequence object, a sequence over its components.  I believe this is the most appropriate interface for frame instances, but if this isn't desired I'm fine with reverting the changes to a normal `getdata()` function or similar.

That's it.

This should fix #710 and #724, and deal with #423 more efficiently.

### DIFF
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -1196,7 +1196,7 @@ cdef class VideoFrame(object):
     cdef readonly Format format
     cdef readonly int width
     cdef readonly int height
-    cdef readonly bint readonly
+    cdef unsigned flags
     cdef readonly VideoProps props
 
     cdef object __weakref__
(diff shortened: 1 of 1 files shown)


myrsloik:
This should probably be targeted at the doodle1 branch and have AudioFrame support as well. At this point I'd rather leave the master branch without big changes. Plus merging this into two places would be super annoying due to how many things I've changed.

myrsloik:
Or actually applying it to both branches makes sense. Would you mind producing the doodle1 branch version as well?

Kamekameha:
Sure, I'll try.  I'm not exactly at all familiar with that branch so it might take a while, though.

By the way, I think now what's left is to deprecate all the getters and planes() from VideoFrame, and the VideoPlane class.  Should I do away with them when I make the changes for doodle1 or are you going to deal with those separately?

myrsloik:
Murder them for doodle1

myrsloik:
This is a friendly reminder poke. I'd like to have this included soon in doodle1 too so I can release a proper RC.

Kamekameha:
Yeah, my bad.  Been working on it, albeit slowly.

PR #729, opened 2021-08-10
core: api4 consistent format names

AkarinVS:
There is a behavior change between v4 and v3 regarding format names:
|  v3  |  v4 |
|-----|----|
|  RGBH  | RGBH48 |
|  RGBS  |  RGBS96 |
| YUV444PH | YUV444PH16 |
| YUV444PS  | YUV444PS32 |

The v4 name does not carry more information as "H"/"S" already unambiguously specifies the length of the components. Furthermore, it confuses downstream tool that parses `vspipe -i` output.

This PR restores the v3 format names.


myrsloik:
Nobody checks that string I hope. I'll remove the superfluous bitness from float formats though since I didn't realize that mistake was in there.

PR #733, opened 2021-08-19
Partially restore CI on windows

AkarinVS:
This PR partially restores CI on windows.

Specifically,
- compiling and testing on x64 is working;
- compiling on x86 is working, testing fails for unknown reasons:
```
ImportError: DLL load failed while importing vapoursynth: A dynamic link library (DLL) initialization routine failed.
```

Replaces previously abandoned effort #717.


AkarinVS:
I'm not able to reproduce the test failure on 32-bit x86 locally.
Probably not worth the effort to track down this issue as we're moving away
from 32-bit VS builds anyway.

myrsloik:
Still doesn't matter since I build everything on windows anyway...

Issue #649, opened 2020-12-31
VideoFrame.get_read_array is somehow able to be written to by Numpy


rlaPHOENiX:
```py
def np_to_frame(array: np.ndarray, frame: vs.VideoFrame) -> vs.VideoFrame:
    for vs_plane, np_plane in enumerate(reversed(range(array.shape[-1]))):
        np.copyto(
            np.asarray(frame.get_read_array(np_plane)),
            array[:, :, vs_plane],
            casting="unsafe"
        )
    return frame
```

Somehow, numpy.copyto is able to write to this frame.get_read_array, as the changes do actually get returned. As you can see, I'm not even using a .copy() of the frame.

This is being called from a ModifyFrame call on a BlankClip with 1 frame length.

myrsloik:
Right, this bug has been known for ages. I completely forgot about it and don't know how to fix it.

WolframRhodium:
The `const` modifier might be required to apply to the memoryview declarations in `vapoursynth.pyx`.

http://docs.cython.org/en/latest/src/userguide/memoryviews.html#read-only-views

PR #735, opened 2021-08-23
Refactor VideoFrame's buffer protocol support [doodle1]

Kamekameha:
This also includes the appropriate changes to AudioFrame.

Because of https://github.com/vapoursynth/vapoursynth/commit/64c1885a0066eb371f437820203ff0a3df99030f the tests in test/expr_test.py now fails.  The fix is simple but I want to submit the change for both doodle1 and master branches.  Unless keeping the two branches in sync in this case is unimportant, then I could just add an extra commit to this PR.

I should also probably add some regression tests for the changes made here, I'll see if I do so later today.

Other than that, these changes are functional and complete.

### DIFF
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -851,7 +851,7 @@ cdef void typedDictToMap(dict ndict, dict atypes, VSMap *inm, VSCore *core, cons
         if val is None:
             continue
 
-        if isinstance(val, (str, bytes, bytearray, VideoNode, VideoFrame)) or not isinstance(val, Iterable):
+        if isinstance(val, (str, bytes, bytearray, VideoNode, RawFrame)) or not isinstance(val, Iterable):
             val = [val]
 
         for v in val:
@@ -1038,7 +1038,7 @@ cdef class FrameProps(object):
         cdef bytes b = name.encode('utf-8')
         cdef const VSAPI *funcs = self.funcs
         val = value
-        if isinstance(val, (str, bytes, bytearray, VideoNode, VideoFrame)):
+        if isinstance(val, (str, bytes, bytearray, VideoNode, RawFrame)):
             val = [val]
         else:
             try:
@@ -1590,6 +1590,30 @@ cdef class AudioFrame(RawFrame):
         cdef int x
         for x in range(self.num_channels):
             yield AudioChannel.__new__(AudioChannel, self, x)
+
+    def __getitem__(self, index):
+        if PyIndex_Check(index):
+            index = PyNumber_Index(index)
+
+            lib = self.funcs
+            frame = <VSFrame*> self.constf
+            format = lib.getAudioFrameFormat(frame)
+
+            if index < 0:
+                index += format.numChannels
+            if not 0 <= index < format.numChannels:
+                raise IndexError("index out of range")
+
+            data = _audio.allocinfo(format)
+            data.base.obj = self
+            data.base.readonly = not self.flags & 1
+
+            _audio.fillinfo(&data.base, frame, index, &self.flags, lib)
+
+            return PyMemoryView_FromObject(data)
+        else:
+            raise TypeError("frame indices must be integers, not %s"
+                            % (type(index).__name__,))
 
     def __len__(self):
         lib = self.funcs
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -1519,9 +1519,6 @@ cdef class AudioFrame(RawFrame):
 
     def __init__(self):
         raise Error('Class cannot be instantiated directly')
-        
-    def __len__(self):
-        return self.funcs.getFrameLength(self.constf)
 
     def copy(self):
         return createAudioFrame(self.funcs.copyFrame(self.constf, self.core), self.funcs, self.core)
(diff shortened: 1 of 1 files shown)


AkarinVS:
I wonder why removing get_{read,write}_array? It's a perfectly fine API.
Also, removing get_read_ptr also breaks Python based previewers.

If we remove unbroken APIs, at least make them go through a proper
deprecation period before removing them.

Migration to api4 is already a big churn for the users, we shouldn't make
it any harder than absolutely necessary.

myrsloik:
I want get_read/write_ptr and get_stride to still exist since those are the absolutely most basic access methods so unhappy users can wrap things however they please. It's a minimal amount of code anyway.
What prevents get_read/write_array from becoming an alias for the new way of accessing things to keep more compatibility? You still return a 2d array of each plane, right?
Btw, why did you add VideoNode and such in this commit? Doesn't seem relevant to the changes. https://github.com/vapoursynth/vapoursynth/pull/735/commits/f58a8ded0c434c4efe9eb026aefd12149262ef5e#diff-b5f2f3a1864e911d9f45d183fd5468dc759abaf0d63a5abb12801cbaeee892a0R854

YomikoR:
Casting the first plane of a frame in RGB48 to a numpy array, with `get_read_array`:
```python
np.asarray(frame.get_read_array(0), dtype=np.uint16)
```
Without:
```python
pt_cast = ctypes.cast(frame.get_read_ptr(0), ctypes.POINTER(ctypes.c_uint8 * frame.get_stride(0) * frame.height))
return np.ctypeslib.as_array(pt_cast, shape=()).view(np.uint16)[:, :frame.width]
```
for YUV with subsampling it will be even more nasty.

Kamekameha:
But why?

None of this is intended to hinder the consumer of the API, most of the same information is accessible through the memoryview returned.  All those removed methods are essentially lower level and ultimately redundant now.  What really changes now is that the data is presented to Python space in a more consistent manner.  If a consumer still wants access to the raw pointers (the only thing not immediately available now) there are always ways around it that shouldn't be any of our concern, like using `PyObject_GetBuffer()` through `ctypes` or even going through numpy's Array Interface.

> If we remove unbroken APIs, at least make them go through a proper deprecation period before removing them.

Except they are indeed broken, two issues were opened recently about it already.  One for reference counting and another for copy-on-write shenanigans.  See #710 and #724.  There's also #649 which this PR fixes too.

> I want get_read/write_ptr and get_stride to still exist since those are the absolutely most basic access methods so unhappy users can wrap things however they please.

<sub>Bruh, you told me I could "murder them", https://github.com/vapoursynth/vapoursynth/pull/727#issuecomment-896607417 (also related https://github.com/vapoursynth/vapoursynth/issues/710#issuecomment-886031635).</sub>

Thing is, the memoryview returned is readily usable, and whatever wrapping a consumer might implement is highly unlikely to look any different.

> What prevents get_read/write_array from becoming an alias for the new way of accessing things to keep more compatibility? You still return a 2d array of each plane, right?

If the concern is compatibility then it'd have to be *fully* compatible, but `memoryview` and `cython.view.memoryview` do not provide the same set of features, I'd have to further wrap Python's `memoryview` into Cython's.  I've already said multiple times by now that I don't agree with returning `cython.view.memoryview` instances to Python space, they've (historically, at least) not been PEP-3118 compliant and they aren't exactly any much more performant.  If consumers want array-like features that Python's `memoryview`s don't provide they should be using numpy instead, which is certainly much more robust than whatever the alternative could be.

> Btw, why did you add VideoNode and such in this commit? Doesn't seem relevant to the changes. https://github.com/vapoursynth/vapoursynth/commit/f58a8ded0c434c4efe9eb026aefd12149262ef5e#diff-b5f2f3a1864e911d9f45d183fd5468dc759abaf0d63a5abb12801cbaeee892a0R854

I didn't?  The diff shows it is **`VideoFrame`** specifically that was added, and I did so because `VideoFrame` (and `AudioFrame`, here) is now a sequence, an iterable, and so it'd would've been iterated over in the process.  Also, as you can see, that's a merge commit of the changes I already made to master, plus solving merge conflicts.  `VideoNode` is there because it was already there;  yes, it should be changed to `RawNode` or something but it's 100% out of the scope of my changes.

Kamekameha:
> Casting the first plane of a frame in RGB48 to a numpy array, with `get_read_array`:
> 
> ```python
> np.asarray(frame.get_read_array(0), dtype=np.uint16)
> ```

With this PR:
```python
np.asarray(frame[0], dtype=np.uint16)
```

There's no way you can tell me this is harder...

DJATOM:
Is it possible yet to write data into frame without `get_write_ptr`? Examples welcome.

Kamekameha:
Yes.
```python
frame = frame.copy()  # make it writable
plane = frame[0]
plane[0, 0] = 0  # plane is a writable buffer
```

DJATOM:
Thanks.

Kamekameha:
I see 516da250420d80997789108195f9fa1f5b7948c6, could've just excluded the last commit from here, the one removing stuff.

Also, no comment on the wall of text I posted?

Anyway, thanks for merging.

AkarinVS:
On Mon, Aug 23, 2021 at 6:35 PM Kamekameha ***@***.***> wrote:

> But why?
>
> None of this is intended to hinder the consumer of the API, most of the
> same information is accessible through the memoryview returned. All those
> removed methods are essentially lower level and ultimately redundant now.
> What really changes now is that the data is presented to Python space in a
> more consistent manner. If a consumer still wants access to the raw
> pointers (the only thing not immediately available now) there are always
> ways around it that shouldn't be any of our concern, like using
> PyObject_GetBuffer() through ctypes or even going through numpy's Array
> Interface.
>
> If we remove unbroken APIs, at least make them go through a proper
> deprecation period before removing them.
>
> Except they are indeed broken, two issues were opened recently about it
> already. One for reference counting and another for copy-on-write
> shenanigans. See #710
> <https://github.com/vapoursynth/vapoursynth/issues/710> and #724
> <https://github.com/vapoursynth/vapoursynth/issues/724>. There's also #649
> <https://github.com/vapoursynth/vapoursynth/issues/649> which this PR
> fixes too.
>
They are not *broken*, just harder to use correctly. All those issues are
known (you need to keep reference to the frames and shouldn't trigger CoW
if you want to use get_read_ptr).
We shouldn't remove an API without first deprecating them as it will break
existing scripts.

Being redundant is not the reason to remove them. We have other redundant
APIs.

myrsloik:
get_read/write_array() were completely broken for any sane definition. It's one thing if a thinly wrapped C-api does unexpected things if you don't follow its rules, it's a very different one when a proper Python object is returned at which point people (like me) assume that it will at least hold a reference to whatever underlying data it's pointing at to avoid access violations. I managed to make that mistake myself. And at that point you may as well use the raw pointer and stop pretending.

PR #738, opened 2021-08-24
cython: restore old low-level frame data access APIs

AkarinVS:
PR #735 removed `get_write_ptr`, `get_read_array` and `get_write_array` on the grounds that they are redundant or broken.

However, a backward compatible API, which will ensure a smooth transition and reduce churn for existing users, far outweighs the added maintenance burden of redundant APIs. Additionally, this PR is also fairly simple, so even the added burden is negligible (besides a moved `class _frame` definition, this PR only changes 11 lines.) 

There is no doubt that the previous implementation of these APIs are buggy and not intuitive, but I believe this new implementation should be fairly consistent.

Specifically, this PR includes three changes:
1. allow `get_read_ptr` on writable frames. If we can write a frame, why can't we read? Also, writable memory view returned from `__getitem__` is also readable, so this restores the symmetry for the raw pointer access API as well.
2. restore `get_write_ptr`. With #735, it's as simple as checking for frame writeability and then return the same pointer as `get_read_ptr`. The underlying mechanism should handle obtaining write pointer correctly while maintaining consistency with memory view returned for the same plane.
3. restore `get_read_array` and `get_write_array` by delegating to `__getitem__`.  Not perfect implementation of the old API, but it should satisfy 99% of the users while also fixing all known bugs (#649, #710, and #724).


Kamekameha:
> [...] on the grounds that they are redundant [...]

To clarify, them being redundant wasn't the primary motivation to remove them, the implementation as of then affected Python objects and I had no intention the "fix" them, I straight up discourage them.  Sure, they'll play along now with the new code but they're also certainly not equivalent.  As a dumb example, take a9122d40fd1a4d1430683b441f1dad434f0fb471, even with this PR the code wouldn't have worked with `get_write_array()` because Python's `memoryview`s can't iterate over multidimensional arrays.

Same non-equivalence for `get_read_ptr()`, users of it might not expect a COW to happen behind the scenes since, you know, the name doesn't imply so.  If anything, this function in particular does not affect the memory views, and such users would've probably expected that this pointer could get invalidated at any point to begin with.

Note that the behavior addressed in your point no. 1 was not introduced by my changes, but the overall phrasing here might make it sound like so.

Other than that, I guess these changes are fine if that makes everybody happy.

src/cython/vapoursynth.pyx
```diff
@@ -1204,6 +1204,23 @@ cdef FrameProps createFrameProps(RawFrame f):
...
+        cdef:
+            unsigned mask = 1 << index+1
+
+        if flags[0] & mask:  # trigger copy-on-write
+            flags[0] &= ~mask  # only do so once, see GH-724
+            return <void*> lib.getWritePtr(frame, index)
+        else:
+            return <void*> lib.getReadPtr(frame, index)
+
+
```
Kamekameha, line 1223: Why moving this around, though?  It seems like unnecessary diff to me, and this commit would be the first thing to appear when git-blame'ing it.  Not like it matters much at the end of the day, but still.

AkarinVS: I was under the assumption it has to be defined before used. apparently that's not necessary, so will revert in the next push.

AkarinVS:
The old APIs are completely restored and I also reverted the expr tests.
The only behavior change is that even getting read ptr/array on a copied
frame will trigger CoW.
I think this is better than the previous behavior and now get_read_ptr ==
get_write_ptr, so there are no surprises.
(I don't think this is a particular important use case anyway, if the user
copied a frame, they should be prepared to write to it.)

PTAL.

Kamekameha (changes requested):
Sigh.  Do not change my code, man.  I rely on Python's `memoryview` deep copy of the Py_buffer passed, Cython's doesn't do that.  I even left a comment here: https://github.com/vapoursynth/vapoursynth/blob/318b802a75ae4de9edef8217949abbde4dce738b/src/cython/vapoursynth.pyx#L1391-L1392
(Admittedly not that descriptive, I guess)

The `_2dview` instance would go out of scope and so its memory, the shape and strides fields.

You could've just wrapped **the result** of `VideoFrame.__getitem__()` instead.

src/cython/vapoursynth.pyx
```diff
@@ -1321,6 +1326,12 @@ cdef class VideoFrame(RawFrame):
         s += '\tHeight: ' + str(self.height) + '\n'
         return s
 
+    def get_read_array(self, int index):
+        return self._getitem_internal(index, True)
+    def get_write_array(self, int index):
+        if self.readonly:
+            raise Error('Cannot obtain write array to read only frame')
+        return self._getitem_internal(index)
```
Kamekameha, line 1334: Here, if you must insist:
```suggestion
    def get_read_array(self, int plane):
        data = VideoFrame.__getitem__(self, plane)
        return view.memoryview(data, PyBUF_RECORDS_RO)

    def get_write_array(self, int plane):
        data = VideoFrame.__getitem__(self, plane)
        return view.memoryview(data, PyBUF_RECORDS)
```

`PyBUF_RECORDS` and `PyBUF_RECORDS_RO` cimported from `cpython.buffer`.

AkarinVS:
Still need to apply change 3 (cython memoryview) to AudioFrame.

Should we also add get_read_array and get_write_array to AudioFrame? I
think this API itself is fine and having a consistent interface seems
desirable.

Kamekameha (changes requested):

Kamekameha (commented):
The old API may be restored but this (https://github.com/vapoursynth/vapoursynth/pull/738/commits/9e1abf7f6382662d54d41d5372b15c03ae713c79) is just noise, specially if/when the old API gets eventually deprecated and removed.

AkarinVS:
If the deep copy of pybuffer is required, why not make it part of _2dview?
Then we can have a much richer API.

AkarinVS:
Anyway, we should add tests.
My test is shown in the 2nd to last commit.

We need tests that cover each subtle corner of the code.

myrsloik:
1. You do realize that the only unreadable frames with the current trivial and safe implementation are frames you, yourself, called copy() on and that would need to read back data you, yourself, wrote. Which is trivial to work around by laundering the frame through a dummy function and by doing so implicitly saying it's immutable and you no longer get to modify your writable frame. 2. I've never seen any usage of get_write_ptr() in the wild. Not a single time.
3. Reasonable if you do this I guess. Big if there.
4. Same main question here, get_write_array() is used where?

Kamekameha:
> If the deep copy of pybuffer is required, why not make it part of _2dview?

Doesn't matter, just use the `memoryview` returned with whichever richer API available, I already suggested a change that still returns your beloved `cython.view.memoryview`s.

AkarinVS:
> 1. You do realize that the only unreadable frames with the current trivial and safe implementation are frames you, yourself, called copy() on and that would need to read back data you, yourself, wrote. Which is trivial to work around by laundering the frame through a dummy function and by doing so implicitly saying it's immutable and you no longer get to modify your writable frame.

i think it's natural to allow read access if you can write, but i do agree this case is not very important.

> 2. I've never seen any usage of get_write_ptr() in the wild. Not a single time.
> 4. Same main question here, get_write_array() is used where?
> 
> And I honestly don't mind small breakage to clean things up. Especially not for the fairly esoteric use cases which I consider this to be.

e.g. https://github.com/WolframRhodium/muvsfunc/blob/35dcd8bd3ed1be785bc0fd2433d9971d6ae983ed/Collections/muvsfunc_numpy.py#L122
and i also have some private python based experimental filters that use get_write_ptr to pass to external libraries (think things like opencv).

this is the only available api if you do prototyping with python based filters.

Kamekameha:
> this is the only available api if you do prototyping with python based filters.

It isn't.  In my PR I suggested "using `PyObject_GetBuffer()` through `ctypes` or even going through numpy's Array Interface".  If you're going to be using numpy anyways then the pointer is at `ndarray.__array_interface__['data'][0]` or, more readable, at `ndarray.ctypes.data`.

AkarinVS:
> > this is the only available api if you do prototyping with python based filters.
> 
> It isn't. In my PR I suggested "using `PyObject_GetBuffer()` through `ctypes` or even going through numpy's Array Interface". If you're going to be using numpy anyways then the pointer is at `ndarray.__array_interface__['data'][0]` or, more readable, at `ndarray.ctypes.data`.

Now that there are two different aspects to this problem:
1. before #735, i think we can all agree that `get_write_{array,ptr}` is the only way to write python based filters.
2. after #735, it's no longer the only way, but we need those APIs for backward compatibility reasons as there are existing scripts that depend on those APIs. Even if we are going to deprecated those, we should at least release one version with both new and old APIs to allow the users to migrate. Otherwise we will be repeating python 2 to python 3 transition.

Kamekameha:
I mean, they still exist in master, with broken behavior and all, along with the new code.

myrsloik:
I'll ponder how to reinsert get_write_ptr() without too many issues at a later time but that's it. If you need 100% compatibility there's always the master branch.

AkarinVS:
PTAL. Now only `get_read_array` and `get_write_array` returns fancy cython memoryviews.

src/cython/vapoursynth.pyx
AkarinVS: Done.

AkarinVS:
> I'll ponder how to reinsert get_write_ptr() without too many issues at a later time but that's it. If you need 100% compatibility there's always the master branch.

I think the `get_write_ptr` part of the PR is not controversial and I have reverted the controversial change to `__getitem__` in the latest push. PTAL.

also, are you suggesting that you will be doing side by side releases of api3 and api4 for an extended time? I was under the assumption that once api4 is stabilized, we will only do a single R55 release based on the doodle1 branch.

myrsloik:
I'll probably keep doing occasional bugfix releases of api3 for a year or so. I don't expect a lot of serious issues to be discovered anyway.

test/expr_test.py
```diff
@@ -4,7 +4,7 @@
 
 def get_pixel_value(clip):
     frame = clip.get_frame(0)
-    arr = frame[0]
+    arr = frame.get_read_array(0)
```
Kamekameha, line 7: As per my comment, https://github.com/vapoursynth/vapoursynth/pull/738#pullrequestreview-738117742, I still see no reason to apply this revert commit.

myrsloik:
Please review my most recent commit that adds back get_write_ptr() and respects the holy COW. With that I consider this issue closed.

PR #739, opened 2021-08-24
std: restore ability to delete frame property with SetFrameProp

### LINKED ISSUE #736: api4: std.SetFrameProp no longer supports deleting frame property
I understand that api4 have introduced `std.RemoveFrameProps` dedicated for this purpose, but I think removing `delete` support from `std.SetFrameProp` breaks backword compatibility for no good reason and makes writing portable scripts that work with both api3 and api4 VS unnecessarily harder than it should be.

What's the rationale behind this backward incompatible change?

Thanks.

AkarinVS:
This PR ensures compatibility with older releases and also recommends the users to use the new and more intuitive `SetFrameProps` and `RemoveFrameProps` APIs.

Fixes #736.

<hr>
The following are copied from [issue comment](https://github.com/vapoursynth/vapoursynth/issues/736#issuecomment-906005013)

The compatibility of releases is important. As is, you can't have a script
that removes frame props working in both R54 and R55 without conditional
code.

And there are code that must work across multiple versions, e.g. mvsfunc:
https://github.com/vxzms/mvsfunc/commit/ca5194854108ae32553decb471595d06e8ee1846

Regrettably, it already contains some conditional code to deal with past
API incompatibilities, but the good news is that most of them are for way
older releases. We managed to keep recent VS releases compatible enough
that most people don't even realize there used to be incompatible changes
on the API.

This is such a simple case: the original API is not broken, just not
intuitive and users have come to depend on them. If we think the newer
alternatives are better, we can certainly advertise new APIs to gradually
phase out existing users. But breaking those existing code is not the most effective way forward.

Could we please make R55 more compatible with R54? (You might think
changing the script to use R55 is acceptable, but in my experience that
makes persuading users to switch significantly harder.)

### DIFF
--- a/doc/functions/setframeprop.rst
+++ b/doc/functions/setframeprop.rst
@@ -1,10 +1,10 @@
 SetFrameProp
 ============
 
-.. function:: SetFrameProp(vnode clip, string prop[, int[] intval, float[] floatval, string[] data])
+.. function:: SetFrameProp(vnode clip, string prop[, bint delete=False, int[] intval, float[] floatval, string[] data])
    :module: std
 
-   Adds a frame property to every frame in *clip*.
+   Adds/deletes a frame property to every frame in *clip*.
 
    If there is already a property with the name *prop* in the frames,
    it will be overwritten.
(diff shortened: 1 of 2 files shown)


AkarinVS:
gentle ping?

PR #740, opened 2021-08-25
std: fix BlankAudio default argument processing

AkarinVS:
According to the docs, explicitly specified arguments override the
one in the provided clip, and also the default length is 10 second,
not 1 hour.


Kamekameha (approved):
I stumbled upon this when testing AudioFrame myself in Python, had to use std.TestAudio instead.

PR #742, opened 2021-08-25
std: fix SplitPlanes

AkarinVS:
ShufflePlanes takes a `planes` argument, not `plane`.


myrsloik:
Do you have secret tests for all functions lying around or do you simply write weird scripts as a hobby?

PR #744, opened 2021-08-26
cython: restore old low-level frame data access APIs (take 2)

AkarinVS:
It seems the reason `get_read_array` and `get_write_array` are removed is because cython's `view.memoryview` is not fully PEP-3118 compliant (e.g. write the buffer to a file when the buffer is not fully contiguous, e.g. #423).

However, this should not be the reason to drop an otherwise working API and break existing user code.

We fix the issue by using a wrapper class that delegates `__getbuffer__` to the newly introduced fully PEP-3118 compliant mechanism from #735, but still exposes the rich interface from cython's `view.memoryview` as in R54 so that no user code will be broken in R55.

All done with minimum amount of added complexity.

Code is tested with an extensive test suite that covers all known use cases.

### DIFF
--- a/doc/pythonreference.rst
+++ b/doc/pythonreference.rst
@@ -159,6 +159,9 @@ associated with it. It is possible to access the raw data using either
 A more Python friendly wrapping is also available where each plane/channel can be accessed
 as a Python array using *frame[plane/channel]*.
 
+For backward compatibility reasons, the VideoFrame class also provides *get_read_array(plane)*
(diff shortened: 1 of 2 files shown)


AkarinVS:
OK, numpy it is.

PTAL, also updated the docs to reflect the backward compatible APIs.
As the API is now fully correct, I don't think we need to deprecate them.

There are more scripts that use these APIs than you think (e.g. all pytorch
based AI filters)

>

PR #767, opened 2021-09-17
Add rpAutoFallback and make it default for API3

misakikasumi:
This commit adds a new request pattern: rpAutoFallback, which means: initially strict spatial by default, fallback to rpGeneral once a non-spatial request is issued.

This commit also makes rpAutoFallback the default request pattern for legacy API3 filters. Before this commit, we treat legacy API3 filters rpGeneral and always add caches for them. However, a large proportion of these filters are actually spatial. By introducing rpAutoFallback, we can treat legacy API3 filters smartly.

The implementation is straightforward: when a non-spatial request occurs, iterate the dependencies of the node, find the corresponding dependency, reset the request pattern to rpGeneral, and then add cache for the dependency node.

### DIFF
--- a/include/VapourSynth4.h
+++ b/include/VapourSynth4.h
@@ -297,7 +297,8 @@ typedef enum VSDataTypeHint {
 typedef enum VSRequestPattern {
     rpGeneral = 0, /* General pattern */
     rpNoFrameReuse = 1, /* When requesting all output frames from the filter no frame will be requested more than once from this input clip, never requests frames beyond the end of the clip */
-    rpStrictSpatial = 2 /* Always (and only) requests frame n from input clip when generating output frame n, never requests frames beyond the end of the clip */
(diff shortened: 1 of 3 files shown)


myrsloik:
It's an interesting idea but I'm not sure it's worth the additional complexity. I have a feeling most common filters will get an api4 build within a few months.
Do you have an example where this actually makes a significant difference?

misakikasumi:
> I have a feeling most common filters will get an api4 build within a few months.

I (and my fellows) *indeed* use some filters that have no newer builds for looooong.

image

misakikasumi:
> It's an interesting idea but I'm not sure it's worth the additional complexity.

I think it has little additional complexity. The fallback detect code is just simple and straightforward. Or we can make rpAutoFallback for internal use only?

> Do you have an example where this actually makes a significant difference?

Theoretically it's certain to further reduce memory usage, since we don't waste space for spatial filters.

misakikasumi:
haf.QTGMC: 700M -> 300M

PR #777, opened 2021-09-22
Rice

sekrit-twc:
Tentative, not tested.

### DIFF
--- a/src/core/vscore.cpp
+++ b/src/core/vscore.cpp
@@ -134,7 +134,7 @@ void VSFunction::call(const VSMap *in, VSMap *out) {
         vs_internal_vsapi.mapSetError(out, "Function was passed values that are unknown to its API version");
         return;
     }
-        
+
     func(in, out, userData, core, getVSAPIInternal(apiMajor));
 }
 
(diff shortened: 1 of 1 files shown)


src/core/vscore.cpp
```diff
@@ -1799,6 +1799,16 @@ VSCore::VSCore(int flags) :
     cpuLevel(INT_MAX),
     memory(new MemoryUse()),
     enableGraphInspection(flags & ccfEnableGraphInspection) {
+#ifdef VS_USE_MIMALLOC
+    static std::once_flag mionce;
+    std::call_once(mionce, []()
+    {
+        mi_option_set_default(mi_option_segment_cache, 32);
+        mi_option_set_enabled_default(mi_option_large_os_pages, 1);
+        mi_option_set_enabled_default(mi_option_eager_commit, 0);
```
misakikasumi, line 1808: Why disable eager commit?

sekrit-twc: In theory, it moves the page fault to the threadpool, preventing (a) serial bottlenecks, and (b) NUMA imbalances, but... Windows doesn't support overcommit, so it doesn't matter.

PR #783, opened 2021-09-25
fix return types

Ichunjo:
We assume here that 'any' is a vnode. Without it, most of the filters will just output a garbage return type


Ichunjo:
Or should I change the return signature in there? https://github.com/Varde-s-Forks/vapoursynth/blob/c8e02a05cb21582a5be935e4894b88548aa014c5/src/core/vsapi.cpp#L43

myrsloik:
You should add the workaround in vsgenstubs. I'd rather not make the any type completely worthless immediately.

PR #792, opened 2021-09-29
Remove gc.collect from every call.

### LINKED ISSUE #785: Performance drop in get_frame_async
```py
clip = core.lsmas.LWLibavSource(path)
# clip[:1000].set_output(0)

import time
start = time.time()
with open(os.devnull, 'wb') as devnull:
    clip[:1000].output(devnull, True)
print()
duration = time.time() - start
print("Duration: ", duration)
```
```
# APIv3
# VideoNode.output
# Duration:  2.153500556945801
# 
# vspipe -p test3.py .
# Output 1000 frames in 3.16 seconds (316.47 fps)



# APIv4
# VideoNode.output
# Duration:  8.244498491287231
# 
# vspipe -p test3.py .
# Output 1000 frames in 2.13 seconds (469.05 fps)
```

VideoNode.output is much slower in api v4 than v3 while the vspipe output is similar minor the performance boost.

cid-chan:
This change seemed to fix the problem.
Can someone confirm?

Fixes #785

### DIFF
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -1583,8 +1583,6 @@ cdef class RawNode(object):
         except Exception as e:
             fut.set_exception(e)
 
-        gc.collect()
-
         return fut
 
     def frames(self, prefetch=None, backlog=None):
@@ -1706,8 +1704,6 @@ cdef class VideoNode(RawNode):
         cdef const VSFrame *f
         self.ensure_valid_frame_number(n)
         
-        gc.collect()
-
         with nogil:
             f = self.funcs.getFrame(n, self.node, errorMsg, 500)
         if f == NULL:
@@ -1961,8 +1957,6 @@ cdef class AudioNode(RawNode):
         cdef const VSFrame *f
         self.ensure_valid_frame_number(n)
         
-        gc.collect()
-
         with nogil:
             f = self.funcs.getFrame(n, self.node, errorMsg, 500)
         if f == NULL:


Ichunjo:
Here are the tests I did:
```py
clip = core.std.BlankClip(width=3840, height=2160, format=vs.YUV420P8, length=2000)

import time
start = time.time()
for f in clip.frames():
    pass

print()
duration = time.time() - start
print("Duration: ", duration)
```
```
R55 APIv3
Duration:  0.875
Core freed but 12441600 bytes still allocated in framebuffers

R55 APIv4
Duration:  3.137500047683716
Core freed but 12441600 bytes still allocated in framebuffers

R56
Duration:  2.3910021781921387

R56 fix
Duration:  1.0894994735717773
```

```py
clip = core.std.BlankClip(width=3840, height=2160, format=vs.YUV420P8, length=5000)
import time
start = time.time()
with open(os.devnull, 'wb') as devnull:
    clip.output(devnull, True)
print()
duration = time.time() - start
print("Duration: ", duration)
```
```
R55 APIv3
Duration:  2.5044965744018555

R55 APIv4
Duration:  7.898499250411987

R56
Duration:  5.440998792648315

R56 fix
Duration:  2.4419970512390137
```
Seems good to me.

DJATOM:
With GC calls - 14.218608140945435
Without GC calls - 13.553486824035645
Tested with plain DGSource in script, maybe suboptimal for APIv4, but apparently there's a difference.

Kamekameha:
Well, now I get huge memory spikes.  IIRC, high RAM usage was the reason for these `gc.collect()`.

Issue #794, opened 2021-09-30
High RAM when the garbage collection of frame-objects is deferred


cid-chan:
Well, now I get huge memory spikes.  IIRC, high RAM usage was the reason for these `gc.collect()` calls.

<sub><sub>Also, I like how off my initial diagnosis of the matter was.  Heh, can't win them all.</sub></sub>

_Originally posted by @Kamekameha in https://github.com/vapoursynth/vapoursynth/issues/792#issuecomment-930597823_

-----

All memory-issues relating to Frame-Instances are almost textbook-examples of python resource-management problems.

RAM are a resource, but usually not that problematic in python applications. For larger applications, file-descriptors are the larger issue, so I will explain the problem using file-descriptors, but the issue the same as the RAM usage of frame-instances.

This stack-overflow answer (https://stackoverflow.com/a/49513137) describes a general problem with garbage collectors -- they do not promise to immediately call a destructor when nobody has a reference on them anymore. Instead they promise to collect them at some point. The reference counting is in fact an implementation detail of CPython many applications unknowingly rely on (see the generator example of https://www.pypy.org/compat.html). This is made worse by cython generating pypy-compatible binaries.

The canonical answer is to allow resources to be closed manually. Resources that have not been closed by the time they are collection will, of course, continue to be closed.

**Here is my proposal:**
We add a `[Video|Audio]Frame.close()`-method and add a context-manager for these frames so users can just do this and never forget to close the resource then:
```py
with clip.get_frame(0) as f:
    ...
```

We add some additional parameter (e.g. `autoclose=False`) to `.frame()` to automatically close the frame after each iteration. The default value is chosen to retain the current behavior of vapoursynth.

`node.output(x)` will then use the above parameter to fix the memory-issues.

**A note on ModifyFrame**
Usually all frames inside the function will not escape the current stack frame. Nor will the stack frame ever be persisted. It should (on CPython at least) continue to be collected then the function call returns to vapoursynth.

Because of this use of frames, we cannot bind a runtime warning to frames that are not manually released at the time of collection. On the other hand, I do not think printing a runtime-warning is neccessary.

myrsloik:
Makes sense. Let's do it.

Kamekameha:
Nah, here's your reference cycle: https://github.com/vapoursynth/vapoursynth/blob/a22068fd0e36a9a2a5c4a6769830868f841b58e5/src/cython/vapoursynth.pyx#L646

[...]

PR #795, opened 2021-10-01
cython: CallbackData.for_future(), avoid a circular reference

### LINKED ISSUE #503: core.max_cache_size is ignored on linux
You can set core.max_cache_size, but Vapoursynth seemingly ignores it, so that the memory usage rises above the threshold without max_cache_size being enforced.

Under Windows, it also seems to be ignored, but only until about "max + 500MB", after which vapoursynth is cleaning up itself.

Arch Linux: 
vapoursynth-git
descale-git

test script:
https://pastebin.com/h0DkxxU4
You can use the test script directly in the Python shell.
Use the process command 2-3 times and your RAM usage should triple the max_cache_size. With each following execution, RAM usage continues to increase.

Kamekameha:
See https://github.com/vapoursynth/vapoursynth/issues/794#issuecomment-931561030.

Not sure if this should be backported to APIv3.

From what I understand this should also fix #503.

### DIFF
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -43,6 +43,7 @@ import weakref
 import atexit
 import contextlib
 import logging
+import functools
 from threading import local as ThreadLocal, Lock, RLock
 from types import MappingProxyType
 from collections import namedtuple
@@ -643,18 +644,20 @@ cdef class CallbackData(object):
     def for_future(self, object future, object wrap_call=None):
         if wrap_call is None:
             wrap_call = lambda func, *args, **kwargs: func(*args, **kwargs)
-        self.callback = self.handle_future
+        self.callback = functools.partial(CallbackData.handle_future,
+                                          self.env, future, wrap_call)
         self.future = future
         self.wrap_cb = wrap_call
 
-    def handle_future(self, node, n, result):
+    @staticmethod
+    def handle_future(env, future, wrapper, node, n, result):
         if isinstance(result, Error):
-            func = self.future.set_exception
+            func = future.set_exception
         else:
-            func = self.future.set_result
+            func = future.set_result
 
-        with use_environment(self.env).use():
-            self.wrap_cb(func, result)
+        with use_environment(env).use():
+            wrapper(func, result)
 
     def receive(self, n, result):
         self.callback(self.node, n, result)


src/cython/vapoursynth.pyx
```diff
@@ -643,18 +644,20 @@ cdef class CallbackData(object):
     def for_future(self, object future, object wrap_call=None):
         if wrap_call is None:
             wrap_call = lambda func, *args, **kwargs: func(*args, **kwargs)
-        self.callback = self.handle_future
+        self.callback = functools.partial(CallbackData.handle_future,
```
cid-chan, line 647: Would it be possible to just rename "handle_future" to "callback"?
Then we would not need to use functools.partial.

myrsloik:
I'm not sure backporting will help much. I removed a lot of circular references in the API4 branch inside the VideoFrame class so it's very likely this won't fix the problem.
But you're free to try.

Kamekameha:
Won't bother then.

src/cython/vapoursynth.pyx
Kamekameha: I don't think so, at least not just a rename.  Scroll a few lines up and you'll notice the callback attribute is not exclusively used to handle just future objects.

Any reason specifically not to use functools.partial?

Kamekameha: Although now that I think about it, what do you think of this?  (I haven't tested it yet.)

```python
    def for_future(self, object future, object wrap_call=None):
        if wrap_call is None:
            wrap_call = lambda func, *args, **kwargs: func(*args, **kwargs)

        def callback(node, n, result, env=self.env):
            if isinstance(result, Error):
                func = future.set_exception
            else:
                func = future.set_result

            with use_environment(env).use():
                wrap_call(func, result)

        self.callback = callback
```

And then we remove the `self.future` and `self.wrap_cb` attributes, and the `handle_future()` method, from the class definition.

Kamekameha:
I noticed that `CallbackData.env` is restored/pushed only for the case of handling a "future" object.  This is only true when the `cb` argument of `RawNode.get_frame_async_raw()` is not callable.

Is this intended?  Shouldn't the environment be pushed regardless?

And, is there actually a use-case for `get_frame_async_raw()` specifically?

cid-chan:
The ides of `get_frame_async_raw` was that it should take any future-object (anything that has `set_exception` and `set_result`), be it from asyncio or some other framework and `wrap_cb` would make sure to make it comply with their threading-model.

cid-chan:
> Is this intended? Shouldn't the environment be pushed regardless?

I think back then CallbackData was also used for output directly. In that case it was not neccessary to push the environment.

src/cython/vapoursynth.pyx
cid-chan: Wouldn't it be a good idea to make it a classmethod?

That way it would be made clear that this is an additional constructor to CallbackData.

myrsloik:
Are we done or do you want to change something?

PR #797, opened 2021-10-04
Automate windows builds

Ichunjo:


myrsloik:
1. Does this really belong here?
2. Bat file for building only works first run, doesn't update the repos and then build so it's really
3. Setting up python is only done once so that bat is probably not that helpful. If you want it to actually do something useful when run again add --upgrade after install.
4. Doesn't automate the whole process. I'd expect it to build docs and then cython module before finishing it off with an installer.
5. I distribute binaries for windows so I'll probably reject this no matter what you do.

Ichunjo:
After I PR'd it, I realised the defects. You're right.

PR #804, opened 2021-10-08
Bump Python version to 3.10

DJATOM:


DJATOM:
Looks like linux autotools had a [bug](https://github.com/amusecode/amuse/blob/main/aclocal.m4#L98) related to python version detection. So that's a reason why linux action fails. Not sure if I can fix that.

DJATOM:
Just wanted to rename branch, but github closed it on it's own. Will make another PR then.

PR #805, opened 2021-10-09
Bump Python version to 3.10

DJATOM:


HolyWu:
> Looks like linux autotools had a [bug](https://github.com/amusecode/amuse/blob/main/aclocal.m4#L98) related to python version detection. So that's a reason why linux action fails. Not sure if I can fix that.

Maybe it's a good time to replace the bloated autotools with meson then.

soldivelot:
https://github.com/vapoursynth/vapoursynth/blob/c80a430b57781216d44519b9fc41192978ad2328/installer/vs-detect-python.bat#L7

DJATOM:
Well, just got up, I went to bed at 5 A.M. Will fix remaining stuff soon.

DJATOM:
Squashed into single commit, makes sense I guess

DJATOM:
```PS C:\Windows\system32> pip install numpy
Collecting numpy
  WARNING: Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'ReadTimeoutError("HTTPSConnectionPool(host='files.pythonhosted.org', port=443): Read timed out. (read timeout=15)")': /packages/3a/be/650f9c091ef71cb01d735775d554e068752d3ff63d7943b26316dc401749/numpy-1.21.2.zip
  Downloading numpy-1.21.2.zip (10.3 MB)
     |████████████████████████████████| 10.3 MB 1.6 MB/s
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
Building wheels for collected packages: numpy
  Building wheel for numpy (PEP 517) ... done
  Created wheel for numpy: filename=numpy-1.21.2-cp310-cp310-win_amd64.whl size=5503044 sha256=f4a52d343046e103130434449835f24d82675e899787ade69c51734a11f31c54
  Stored in directory: c:\users\djatom\appdata\local\pip\cache\wheels\55\ca\9f\48d8bbedba076d0949e6d329c6aad3705267c63cc38521852d
Successfully built numpy
Installing collected packages: numpy
Successfully installed numpy-1.21.2```
Hmm, but it manages to complete installation. Where to spot incompatibilities?

DJATOM:
Okay, I'll let Myrsloik decide then. If there will be new incompatibilities against further commits, I'll resolve those on sight.

myrsloik:
> Not sure if it's intentional, but this seems to break installing VS on Python 3.9 over all. Is there any reason to support 3.8 and 3.10, but not 3.9?

I generally only support the latest usable python version (usable defined as not too new so all popular packages have been updated). I added 3.8 in addition to that since it's the last win7 compatible version.

Supporting more than 2 versions at once is also very annoying. Python 3.9 has no interesting property and won't be supported in the long run.

PR #815, opened 2021-10-12
Improve get_frame async raw and remove a long deprecated function

cid-chan:
1. Remove the deprecated function vpy_current_environment()
2. Refactor get_frame_async and deprecate get_frame_async_raw to be easier to understand for the user. It also simplifies the code of `CallbackData`.


myrsloik:
Why do the tests fail?

PR #818, opened 2021-10-13
Update python reference doc

Ichunjo:
`add_log_handler` and `remove_log_handler` descriptions have been copied from the C API reference.

### DIFF
--- a/doc/pythonreference.rst
+++ b/doc/pythonreference.rst
@@ -16,7 +16,7 @@ special things unique to Python scripting, such as slicing and output.
 VapourSynth Structure
 #####################
 
-Most operations in the VapourSynth library are performed through the singleton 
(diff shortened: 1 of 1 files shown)


doc/pythonreference.rst
```diff
@@ -704,18 +714,18 @@ Classes and Functions
...
 
       A similar problem also existed in previous VapourSynth versions!
-      
+
       .. code::
 
-         env = vpy_current_environment()
+         env = get_current_environment()
          with env:
               yield
```
cid-chan, line 730: `vpy_current_environment` has this problem. `get_current_environment` cannot be directly used as a context manager.

Ichunjo: Indeed, sorry!

PR #841, opened 2022-01-11
vsscript: Fix loading python module in mingw

Biswa96:
Search for portable python library only if compiled with
MSVC toolchain. In MINGW, shared libraries are already
linked with libpython*.dll


myrsloik:
Mingw isn't a supported build configuration for windows

PR #871, opened 2022-05-27
Add an environment variable VAPOURSYNTH_CONF_PATH

cid-chan:
This allows a user to provide a custom vapoursynth.conf.

### DIFF
--- a/src/core/vscore.cpp
+++ b/src/core/vscore.cpp
@@ -1858,21 +1858,33 @@ VSCore::VSCore(int flags) :
 
 #else
     std::string configFile;
-    const char *home = getenv("HOME");
 #ifdef VS_TARGET_OS_DARWIN
-    std::string filter = ".dylib";
-    if (home) {
-        configFile.append(home).append("/Library/Application Support/VapourSynth/vapoursynth.conf");
-    }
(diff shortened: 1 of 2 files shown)


src/core/vscore.cpp
```diff
@@ -1858,21 +1858,33 @@ VSCore::VSCore(int flags) :
...
+        }
 #else
-    std::string filter = ".so";
-    const char *xdg_config_home = getenv("XDG_CONFIG_HOME");
-    if (xdg_config_home) {
-        configFile.append(xdg_config_home).append("/vapoursynth/vapoursynth.conf");
-    } else if (home) {
-        configFile.append(home).append("/.config/vapoursynth/vapoursynth.conf");
-    } // If neither exists, an empty string will do.
+        std::string filter = ".so";
```
FichteFoll, line 1879: `filter` is already declared and defined in line 1862 and 1864 respectively.

cid-chan: I'm not that well versed in C++, so I kept like it was before as I was uncomfortable with refactoring it.

cid-chan:
I think I managed to find the one who kept closing this PR.

PR #882, opened 2022-06-19
macos.yml: Run on macOS 12

HolyWu:


cid-chan:
Why isn't macos-latest not enough?

HolyWu:
I simply happened to find out that using macos-12 mysteriously solved https://github.com/vapoursynth/vapoursynth/runs/6950495478?check_suite_focus=true. Alternatively you may pass `/usr/local/lib` as additional directory to be searched for libvapoursynth. I don't really care, anyway.

PR #889, opened 2022-07-18
Log a warning if autoloading a plugin fails

Frechdachs:
As of now, failing to autoload a plugin is just silently ignored.
Adding a warning may annoy some people that have a messy plugin directory, but that's a good thing.
This warning helped me find some broken plugins.


theChaosCoder:
I usually loop over all dlls with LoadPlugin(). This would simplify things.

Issue #888, opened 2022-07-12
vapoursynth-portable (pypi) path issue


YomikoR:
After running `pip install vapoursynth-portable` in Windows, vsscript.dll is in the Scripts folder, where, the required python310.dll cannot be found.

A possible workaround may be reading the relative/absolute path of python dll from portable.vs when in portable mode.

myrsloik:
Only the installer is officially supported

PR #891, opened 2022-08-04
Let vsscript search python dll in parent dir as well in portable mode

YomikoR:
Partially solves #888 when using python 3.10 and when not in a virtual environment.


myrsloik:
Why is this specific path relevant?

YomikoR:
> Why is this specific path relevant?

portable.vs detection of vsscript.dll is not documented. People usually get it either from a portable VS release (works) or from vapoursynth-portable package (see below).
https://github.com/vapoursynth/vapoursynth/blob/master/installer/setup.py#L75

Issue #893, opened 2022-08-14
Cannot recover from arError within a filter.


cid-chan:
## Original Problem:
Within a `FrameEval` and `ModifyFrame`, you cannot recover from errors that are caused during any clips passed in 
`FrameEval(prop_src)`/`ModifyFrame(clips)`.

This is especially devastating, if this happens within a chain of `FrameEval`s, as this could easily become resource leak.

To solve this problem, I propose the following changes:

`VSAPI`

> **1a.** Add a function, `const char* getFilterError(VSFrameContext* ctx)`, that takes the current FrameContext and returns the error message.
> The frame context retains ownership of the error.
> 
> Pro:
> - You can find out the error during `arError`
> 
> Con:
> - You don't exactly know what frame the error came from.
> 
> **1b.** (Adapted from Myrsloiks idea) Instead or in both cases, add function `const VSMap* getFrameFilterMaybe(VSFrameContext* ctx)` that returns a map with `frame` if the frame was successful, or has an error set.
> 
> Con:
> - I do not know how much you need to change in the current code.
> Pro:
> - You can find out which filter caused an error.
> 
> **2a.** In `setFilterError(const char* error, VSFrameContext* ctx)`, remove the `nullptr` check.Passing `nullptr` as error, should clear the error from the ctx.
> Pro:
> - Doesn't add a new function
> Con:
> - Might be more error-prone to the developer
> 
> **2b.** Or, instead of 2a, add a function `clearFilterError(VSFrameContext* ctx)` which clears any error from ctx.
> Pro:
> - Clearly communicates to vapoursynth that the develop really does wish to clear the filter error.
> Con:
> - Requires adding a new function.

Solution 1:

> `std.ModifyFrame/std.FrameEval` 
> Append a new optional parameter to both functions. This parameter is a callback that is called when `arError` is set as the activation reason.
> Pro:
> - No need to a new filter.
> 
> Con:
> - If you requested more than one frame, it is not possible to find out which of them errored out.
> - If multiple errors occur, they are all shadowed.
> 

Solution 2:

> Add a new filter `std.Recover(clip: clip, recover: func)`.
> All requests for frames should be forwarded to clip unchanged.
> If `arError` occurs, `func` should be called with the error message as its only argument.
> `recover` **must** return a supported frame.
> If `recover` errors, it should shadow the previous error.
> Optionally, it could replace the error with `"{shadowed_error}\n\nDuring recovery of this error, this error occured: {new_error}"`
[...]

PR #894, opened 2022-08-14
PoC #893 solution 2

cid-chan:
Implements #893


myrsloik:
Probably won't happen.

PR #901, opened 2022-09-18
Update VideoFrame's `writelines` method

Setsugennoao:
Complementary PR of vsrepo's, so that this method can be added to public typing stubs.

EDIT: Following cid's request, I converted this PR into updating this method, so it's more clear and pythonic; instead of taking a callback, it now yields chunks of data.

The only public use I know of is in [lvsfunc](https://github.com/Irrational-Encoding-Wizardry/lvsfunc/blob/master/lvsfunc/render.py#L65), and for that @tomato39 had to [modify](https://github.com/tomato39/vsrepo/commit/318ac44e64f423fade47517bf1bc70da09c1f234) the template.
CC: @LightArrowsEXE, you'll need to update that line.

_Originally posted by @cid-chan in https://github.com/vapoursynth/vsrepo/pull/186#discussion_r973683694_

### DIFF
--- a/doc/pythonreference.rst
+++ b/doc/pythonreference.rst
@@ -474,7 +474,7 @@ Classes and Functions
 
       Returns the stride between lines in a *plane*.
 
-   .. py:method:: writelines()
+   .. py:method:: writechunks()
 
       This method is usually used to dump the contents of a VideoFrame to disk.
       The returned generator yields contiguous chunks of the VideoFrame memory.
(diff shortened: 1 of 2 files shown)


cid-chan:
Can you please add documentation to it?

cid-chan:
I think the API is good. But shouldn't it be called `chunks` or similar?

cid-chan (approved):

PR #908, opened 2022-10-15
Fix ownership not transferred properly from unique_ptr

Endilll:
```
foo(u_ptr.get());
u_ptr.release();
```
is a roundabout way to transfer ownership. And a more lengthy one.
Ideally `release()` should be marked `[[nodiscard]]`, because it's not meant to be called just for its side effects, but it's missing, as well as in many-many other places in the standard library.


Endilll:
Well, getting data via `unique_ptr` and releasing the ownership in the same function call doesn't work too good together.

PR #911, opened 2022-11-16
Update build scripts

Setsugennoao:
Update build scripts to be easier to use for developers.

* Add single `compile_all.bat`
* Add `install_deps.bat` which clones dependencies and on sequential runs pulls new commits (gets called by `compile_all.bat`)
* Add and check multiple install paths other than `C:`
* Fix py.exe / py ambiguity (yes, with multiple installs it breaks...)
* Add vsgenstubs4 to the portable version, alongside vsrepo.

Potentially breaking changes
* Rename `make portable.bat` to `make_portable.bat`
* Remove `doc/Makefile` (substitued with sphinx commands)


cid-chan:
I do not own a windows system anymore. I cannot comment on this PR.

RivenSkaye:
I tested the build process. Everything seems to work quite nicely, though the updated scripts seem to be looking for Visual Studio 2022 rather than 2019 as the documentation lists. Docs would need to be reworked for this for sure though.

If I'm not missing anything, this only needs the Python header files, so would it be possible for the automatic script to also pull in the required tarballs off the Python [dot] org archives and unpack them locally using whatever 7-zip it grabs? That would prevent the need for people to install 4 concurrent Python versions (because now I'm stuck uninstalling 3)

Setsugennoao:
> If I'm not missing anything, this only needs the Python header files, so would it be possible for the automatic script to also pull in the required tarballs off the Python [dot] org archives and unpack them locally using whatever 7-zip it grabs? That would prevent the need for people to install 4 concurrent Python versions (because now I'm stuck uninstalling 3)

I personally don't like the idea, plus that's assuming you only want to compile the vsscript project.
In reality you 110% will also compile the cython extention and that requires the corresponding python version with all the dependencies installed to compile.

.gitmodules
```diff
@@ -0,0 +1,12 @@
...
+	url = https://github.com/AviSynth/AviSynthPlus.git
+[submodule "libp2p"]
+	path = libp2p
+	url = https://github.com/sekrit-twc/libp2p.git
+[submodule "vsrepo"]
+	path = vsrepo
+	url = https://github.com/vapoursynth/vsrepo.git
+[submodule "zimg"]
+	path = zimg
+	url = https://github.com/sekrit-twc/zimg.git
```
cid-chan, line 12: Iirc, Myr used this once, and was dissatisfied.
I don't like the feature either. You're gonna have to find another solution for that.

cid-chan (changes requested):

.gitmodules
Setsugennoao: This isn't a solution to anything, the build scripts grab em from elsewhere, it's just that i think it's just nornal to link to other repos in github too.
If myrsloik doesn't like it for w/e reason we can just delete the file

PR #918, opened 2022-12-15
Update VS classes __str__/__repr__

Setsugennoao:
Added missing members, or whole repr in the case of AudioFrame, also made all consistent between each other.

### DIFF
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -2766,10 +2766,12 @@ cdef class Function(object):
         raise Error('Class cannot be instantiated directly')
 
     cdef is_video_injectable(self):
-        return self.signature.find(':vnode') > 0
-        
+        first_arg_i = self.signature.find(':')
+        return first_arg_i > 0 and self.signature.find(':vnode') == first_arg_i
+ 
     cdef is_audio_injectable(self):
-        return self.signature.find(':anode') > 0
+        first_arg_i = self.signature.find(':')
+        return first_arg_i > 0 and self.signature.find(':anode') == first_arg_i
 
     def __call__(self, *args, **kwargs):
         cdef VSMap *inm
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -1629,11 +1629,7 @@ cdef class AudioFrame(RawFrame):
         return lib.getAudioFrameFormat(self.constf).numChannels
 
     def __str__(self):
-        channels = ', '.join([
-            AudioChannels(v).name
-            for v in AudioChannels
-            if ((1 << v) & self.channel_layout)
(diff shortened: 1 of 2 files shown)


cid-chan:
Well. Since you made the PR, I have to break the bad news to you: I always wanted to switch to a single line-repr since this is what most people expect.

src/cython/vapoursynth.pyx
```diff
@@ -2435,6 +2459,9 @@ cdef class _CoreProxy(object):
         
     def __setattr__(self, name, value):
         setattr(self.core, name, value)
+
+    def __str__(self):
```
cid-chan, line 2463: We need a case if there is no environment. Otherwise `__str__` crashes.

cid-chan (changes requested):

myrsloik:
I kinda miss the num channels bit in AudioNode string representation. The rest looks ok apart from cid's comment.

Setsugennoao:
This now depends on #919

Setsugennoao:
To add `Plugin` and `Function`'s `__repr__`, I opened #920 to add `version`/`bound` parameters.

Setsugennoao:
After using this branch for a couple of days, I'm _really_ preferring the repr over the str...
At this point I'd just delete the old __str\_\_ formatting and only leave the new __repr\_\_, lol.

src/cython/vapoursynth.pyx
```diff
@@ -1225,6 +1267,47 @@ cdef FrameProps createFrameProps(RawFrame f):
 Mapping.register(FrameProps)
 
 
+cdef class ChannelLayout(int):
+    cdef readonly int num_channels
```
cid-chan, line 1271: Not required.

src/cython/vapoursynth.pyx
```diff
@@ -1225,6 +1267,47 @@ cdef FrameProps createFrameProps(RawFrame f):
...
+            if ((1 << v) & self):
+                yield v
+
+    def __eq__(self, other):
+        if not isinstance(other, ChannelLayout):
+            return False
+        return other == self
+
+    def __len__(self):
+        return self.num_channels
```
cid-chan, line 1290: I think, `self.bit_count()` would be another solution that doesn't require `num_channels`.

cid-chan (changes requested):

src/cython/vapoursynth.pyx
Setsugennoao: All right, so we only use the channel layout and can put this on public API

src/cython/vapoursynth.pyx
Setsugennoao: All right, so we only use the channel layout and can put this on public API

PR #919, opened 2022-12-15
Add channels properties to AudioNode/AudioFrame (ChannelLayout)

Setsugennoao:
As discussed on discord.

### DIFF
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -1225,12 +1225,7 @@ cdef FrameProps createFrameProps(RawFrame f):
 Mapping.register(FrameProps)
 
 
-cdef class ChannelLayout(int):
-    cdef readonly int num_channels
-
-    def __init__(self):
-        raise Error('Class cannot be instantiated directly')
-
+class ChannelLayout(int):
     def __contains__(self, layout):
         return bool(self & (1 << layout))
 
@@ -1245,7 +1240,7 @@ cdef class ChannelLayout(int):
         return other == self
 
     def __len__(self):
-        return self.num_channels
+        return self.bit_count()
 
     def __repr__(self):
         cls = self.__class__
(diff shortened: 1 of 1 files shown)


Setsugennoao:
Should we consider moving the createChannelLayout to createAudioNode/createAudioFrame and make channels an effective property instead of going through the `property` descriptor?
It just inherits `int`, I think the overhead will be negligible.

src/cython/vapoursynth.pyx
```diff
@@ -1225,6 +1225,46 @@ cdef FrameProps createFrameProps(RawFrame f):
 Mapping.register(FrameProps)
 
 
+cdef class ChannelLayout(int):
```
cid-chan, line 1228: That doesn't need to be cdef and couold be instantiated directly. Indeed, I wonder if this might be a thing, that people might want to do.

I mean it's just an int at the end of the day. It doesn't contain pointers to vapoursynth.

cid-chan (changes requested):

src/cython/vapoursynth.pyx
Setsugennoao: I just didn't want people to use it directly, but if you want it as part of the public API, sure.

src/cython/vapoursynth.pyx
```diff
@@ -1225,6 +1225,36 @@ cdef FrameProps createFrameProps(RawFrame f):
 Mapping.register(FrameProps)
 
 
+class ChannelLayout(int):
+    def __contains__(self, layout):
+        return bool(self & (1 << layout))
+
+    def __iter__(self):
```
cid-chan, line 1232: Replace this function with an appropriate `__getattr__`, and python will automatically write the following `__iter__` implementation for you:

```py
def __iter__(self):
    for i in range(len(self)):
        yield self[i]
```

PR #920, opened 2022-12-20
Update Plugin/Function

Setsugennoao:
- Add `Plugin.__version__` with `PluginVersion` tuple.
 - Fix `Function.is_video/audio_injectable` assuming the first argument of an injectable function will have `:v/anode`.

### DIFF
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -2465,7 +2465,9 @@ cdef class _CoreProxy(object):
         setattr(self.core, name, value)
     
 core = _CoreProxy.__new__(_CoreProxy)
-    
+
+PluginVersion = namedtuple("PluginVersion", "major minor")
+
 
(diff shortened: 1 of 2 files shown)


src/cython/vapoursynth.pyx
```diff
@@ -2496,6 +2498,15 @@ cdef class Plugin(object):
             func = self.funcs.getNextPluginFunction(func, self.plugin)
             yield tmp
 
+    @property
+    def __version__(self):
```
cid-chan, line 2502: Remove the dunder.

cid-chan (changes requested):

src/cython/vapoursynth.pyx
Setsugennoao: Reason i added the dunder is some packages actually use version/Version as name for functions, and i didn't want to break those.

PR #953, opened 2023-05-13
Update workflows

HolyWu:
https://github.com/vapoursynth/vapoursynth/commit/1239124c2d70f32c468326e48bfa7e9362550c36 changed to use `std::from_chars` from C++17. Unfortunately LLVM's libc++ [does not yet implement](https://libcxx.llvm.org/Status/Cxx17.html#note-p0067) `std::from_chars for float and double`. Hence it breaks compilation on Clang/LLVM with libc++.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -7,26 +7,11 @@ jobs:
     runs-on: ubuntu-latest
 
     env:
-      CC: ${{ matrix.cc }}
-      CXX: ${{ matrix.cxx }}
-
-    strategy:
-      matrix:
-        include:
-          - cc: gcc-11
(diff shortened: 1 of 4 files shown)


Setsugennoao:
Why does this PR mix workflow, typos, and PresetVideoFormat?

HolyWu:
They are meant to fix failed CI.

Setsugennoao:
Can you also rename the struct in the docs?

HolyWu:
I missed that. Rebased.

myrsloik:
Will merge if you split it out into ONLY the workflow bits

HolyWu:
Done.

PR #966, opened 2023-06-22
Add mtDeprecation and annoying warnings for plugins.

### LINKED ISSUE #912: Version numbers in plugins are usually 0, -1 or similar
Solution: add an annoying warning whenever loading one of these plugins

Setsugennoao:
As discussed on discord [here](https://discord.com/channels/856381934052704266/856406641872207903/1115631205166108723) and for code cleanness, and for being able to distinguish the logs I added mtDeprecation (idea from a comment by @YomikoR [here](https://github.com/vapoursynth/vapoursynth/issues/912#issuecomment-1410513304)).

- Resolves #912, then in the future (with APIv3 drop ~next year?) the warning can be swapped with an `API misuse` fatal error.
- Starts warning users of plugins still on APIv3.

### DIFF
--- a/doc/api/vapoursynth4.h.rst
+++ b/doc/api/vapoursynth4.h.rst
@@ -614,6 +614,8 @@ enum VSMessageType
 
    * mtFatal
 
+   * mtDeprecation
+
 
 .. _VSCoreCreationFlags:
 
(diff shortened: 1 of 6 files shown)


myrsloik:
The point is to annoy people a bit. And you actually change the ABI so all editors would have to be updated. Or the have the messages secretly remapped to a warning again. Go open issues

PR #968, opened 2023-06-22
Fix BlankClip deliveredInfo structure

Setsugennoao:
VSVideoInfo with dynamic format should also have Integer sample type as specified here
https://github.com/vapoursynth/vapoursynth/blob/master/src/core/vscore.cpp#L1318-L1319

Can reproduce with: `core.std.BlankClip(format=vs.GRAYS, varformat=True)`
```
vapoursynth.Error: The VSVideoInfo structure passed by BlankClip is invalid.
```


myrsloik:
Actually just doing format = {} is the intended way if you don't brainfart, the representation was chosen that way for simplicity

PR #970, opened 2023-06-22
Fix typo VS_TARGET_OS_WINDOWS => VS_TARGET_CPU_X86

Setsugennoao:
`x86utils.h` will only be imported when `VS_TARGET_CPU_X86`, and of course `vs_isSSEStateOk` can only be used when compiling with it.


myrsloik:
I guess, however I doubt that file will ever be relevant outside windows+x86 since it's for binary compatibility

PR #971, opened 2023-06-22
Add meson build system

Setsugennoao:
Doesn't remove any existing workflows, just adds the new build system, with the integrations from #964.

Based on #825, updated and with docs additions.

Let me know if I should also add Linux/MacOS/MinGW workflows.

### DIFF
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -154,7 +154,13 @@ Linux and OS X Compilation
 **************************
 
 These are the requirements:
-   * Autoconf, Automake, and Libtool, probably recent versions
+   - With make:
+      * Autoconf, Automake
+      * Libtool
+
+   - With meson-ninja:
+      * Meson 0.60.0 or later
+      * ninja-buld
 
    * pkg-config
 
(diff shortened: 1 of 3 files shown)


diizzyy:
That looks really good! For packaging purposes it would be very nice if you could add a compile test/check for avx2 during configure stage and add a switch to force it disabled.

myrsloik:
Nope

diizzyy:
Out of curiosity, any reasons why?
Not saying that GNU Autotools will go away anytime soon but upstream have expressed concerns [1] about the future and many projects in general are migrating to either CMake and Meson or already have.

https://lists.gnu.org/archive/html/autoconf/2021-01/msg00049.html

myrsloik:
> Out of curiosity, any reasons why? Not saying that GNU Autotools will go away anytime soon but upstream have expressed concerns [1] about the future and many projects in general are migrating to either CMake and Meson or already have.
> 
> https://lists.gnu.org/archive/html/autoconf/2021-01/msg00049.html

1. The current solution works
2. Not discussed with me at all before doing actual work
3. User known to provide PRs with very little testing
4. Doesn't rip out autotools so now it's twice the work and I hate build systems the most of all things in this world
5. In the end I will have to fix all small problems that show up, I have very little of at the moment, the master branch will then simply languish in an uncompilable state for months

At some point it may be changed but I'm not in a hurry. Come back when autotools actually creates problems. You just gotta use the nope card liberally to keep things managable.

Setsugennoao:
I did at least test this! (linux server, wsl)

PR #973, opened 2023-06-24
Fix #cab48c4 (32 bits removal)

Setsugennoao:


myrsloik:
I don't want to remove the actual visual studio project support, the reason is very simple: it helps to find things that generate warnings in 32 bit mode. So the first 2 commits are denied. The rest is fine I guess.

Setsugennoao:
You can only compile one type at a time, so you should set the 32 bits python to `VSPYTHON38_PATH`, doesn't make sense to set multiple if you're doing it manually, no?
No other part of the process sets `VSPYTHON38_PATH`, and it'd be counter-intuitive to have a different include variable just for that.

Setsugennoao:
Though I guess you could make an argument for the fact that I made it automatic, and you don't want to lose it even for 32bits?

myrsloik:
> Though I guess you could make an argument for the fact that I made it automatic, and you don't want to lose it even for 32bits?

If it's just the automatic bit it can go

myrsloik:
The "Fix 32bits registry paths" commit has to go too. I don't want to change the registry paths if someone actually bothers to keep making compiles.

PR #1004, opened 2023-11-21
Update Core properties in cython module

### LINKED ISSUE #1003: Expose all information from getCoreInfo in python
Currently used memory isn't exposed

Setsugennoao:
This PR mainly updates the last functions that should be properties (getters).

`version` is useless as it was superseded by the new `__str__` implementation that formats `CoreInfo.versionString`.

`version_number` was also ambiguous, so I added `core_version` to remove the ambiguity and `api_version` because that never got exposed other than in `version()` but you had to parse the string.

Those getters now return the version namedtuples we use in the module (`__version__`, `__api_version__`) to be consistent. 
About those namedtuples, I also added a `__str__` to them so they get formatted as everywhere else in the application. (Rxx/Rxx.y)

Then I added `used_cache_size` which is a simple getter to return CoreInfo.usedFramebufferSize (Fixes #1003 )

### DIFF
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -83,8 +83,25 @@ __all__ = [
   'core',
 ]
 
-__version__ = namedtuple("VapourSynthVersion", "release_major release_minor")(VS_CURRENT_RELEASE, 0)
-__api_version__ = namedtuple("VapourSynthAPIVersion", "api_major api_minor")(VAPOURSYNTH_API_MAJOR, VAPOURSYNTH_API_MINOR)
+class VapourSynthVersion(typing.NamedTuple):
+    release_major: int
+    release_minor: int
+
+    def __str__(self):
+        if self.release_minor:
+            return f'R{self.release_major}.{self.release_minor}'
+        return f'R{self.release_major}'
+
+class VapourSynthAPIVersion(typing.NamedTuple):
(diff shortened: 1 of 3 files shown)


Setsugennoao:
Fun fact: those functions were added _almost exactly_ 10 years ago :o https://github.com/vapoursynth/vapoursynth/commit/21148b9732b96ffadd4066599f1b8d10fc51b94a

myrsloik:
Update the docs too and I'll merge it

Setsugennoao:
@myrsloik should we consider returning in MiB as float? For example, you can't know if the cache is really empty or there's, for example, 100KiB 🤔

src/cython/vapoursynth.pyx
```diff
@@ -2495,6 +2495,12 @@ cdef class Core(object):
             new_size = new_size * 1024 * 1024
             self.funcs.setMaxCacheSize(new_size, self.core)
 
+    @property
+    def used_cache_size(self):
+        cdef VSCoreInfo v
+        self.funcs.getCoreInfo(self.core, &v)
+        return <int64_t>v.usedFramebufferSize
```
myrsloik, line 2502: I don't think a cast is needed here.

Setsugennoao: Fixed

PR #1009, opened 2023-12-13
Update cython to expose latest API changes

Setsugennoao:
- Fixed (very) minor typos (that's why i included them here)
- `getCoreNodeTiming` added in C++ API as asked on Discord (for making a getter/setter timings.enabled instead of just having a set_enabled or smth function, which would be weird as everything else is a property)
  let me know if this isn't wanted or maybe should add a flag in setCoreNodeTiming to just return the current state
- properly exposed the now stable APIs for inspecting node properties
  - these were also behind the core creation flags but now don't require any check in the C++ side so I took the liberty of removing it from python too, let me know if you want to still keep it behind the ccf (which would be a bummer since pretty much it will only work through vspipe and previewers)
- added a simple wrapper for the timing stuff of core, and added the possibility to reset the counter for the RawNode timings too

Will make a PR for vsgenstubs too once this PR gets reviewed.


src/cython/vapoursynth.pyx
```diff
@@ -1956,9 +1956,15 @@ cdef class RawNode(object):
...
         return self.funcs.getNodeName(self.node).decode("utf-8")
 
-    @property
-    def timings(self):
-        return self.funcs.getNodeProcessingTime(self.node, 0)
+    property timings:
+        def __get__(self):
+            return self.funcs.getNodeProcessingTime(self.node, False)
+        
+        def __set__(self, int value):
```
myrsloik, line 1963: why not "bint reset" if all you can do is pass 0 anyway?

Setsugennoao: No particular reason other than int is used in VSAPI too, I'll change it.

PR #1010, opened 2023-12-13
Make error buffer size bigger in cython get_frame

Setsugennoao:
I had already talked about this feature to @cid-chan a couple months ago.

If you have a better way to implement this I'd be happy to implement it but for now I just added a variable that's in EnvironmentData so it's pretty inaccessible to normal users.

Main application is previewers really, since they use get_frame to display frames and when there's errors in frame evals and such they always get truncated.

I also bumped the default characters from 512 to 1024 since pretty much everything got truncated; let me know if I should revert this.

### DIFF
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -2045,13 +2045,13 @@ cdef class VideoNode(RawNode):
             raise ValueError('Requesting frame number is beyond the last frame')
 
     def get_frame(self, int n):
-        cdef char errorMsg[512]
+        cdef char errorMsg[4096]
         cdef char *ep = errorMsg
         cdef const VSFrame *f
         self.ensure_valid_frame_number(n)
 
         with nogil:
-            f = self.funcs.getFrame(n, self.node, errorMsg, 500)
+            f = self.funcs.getFrame(n, self.node, errorMsg, 4096)
         if f == NULL:
             if (errorMsg[0]):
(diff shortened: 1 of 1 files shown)


myrsloik:
It's on the python side and only in scripts where you get individual frames so the allocation is truly insignificantly small. I'd bump it to 4kb and skip configurability. If someone has a longer error message they should write better code.

Setsugennoao:
```py
import time
import vapoursynth as vs

rounds = 1000
clip = vs.core.std.BlankClip(None, 1, 1, vs.GRAY8)


def x():
    clip.get_frame(0)

    start = time.perf_counter()

    for _ in range(10000):
        clip.get_frame(0)

    end = time.perf_counter()

    return end - start


print(sum(x() for _ in range(rounds)) / rounds)
```


```
benchmark of 10000 iterations of get_frame hitting cache, so ideally just cython overhead (in seconds)
static 512   => 0.14154506129938818
static 4096  => 0.17828076960005274 (20% slower, ~40ms more over a baseline 140ms)
dynamic 512  => 0.14457755119996726 (2% slower, ~3ms more for the malloc call)
dynamic 4096 => 0.18202331540011800 (20% slower, ~40ms more over a baseline 144ms)
```

TL;DR making the buffer 4096 makes it
0.000000003673570830066455212743 *milliseconds* slower per iteration

guess I'll just change it

PR #1013, opened 2023-12-17
Add prop_src to ShufflePlanes/SplitPlanes

### LINKED ISSUE #844: Add an option to remove matrix property in SplitPlanes function?
When I'm doing `SplitPlanes` and convert the output clips to YUV, there's an error complaining about the GREY clips cannot be having a RGB matrix property. I would have to manually call `RemoveFrameProps` to remove it. Would it be better to just have this option within the `SplitPlanes` function?

Setsugennoao:
Fix #844

Implements https://github.com/vapoursynth/vapoursynth/issues/844#issuecomment-1685313650

In this PR `ShufflePlanes` and `SplitPlanes` take a `prop_src` parameter that can be an empty array or the number of planes they output (aka ShufflePlanes only takes one and SplitPlanes one or three).

### DIFF
--- a/doc/functions/video/shuffleplanes.rst
+++ b/doc/functions/video/shuffleplanes.rst
@@ -1,7 +1,7 @@
 ShufflePlanes
 =============
 
-.. function::   ShufflePlanes(vnode[] clips, int[] planes, int colorfamily)
+.. function::   ShufflePlanes(vnode[] clips, int[] planes, int colorfamily[, vnode[] prop_src=clips[0]])
    :module: std
 
    ShufflePlanes can extract and combine planes from different clips in the most
@@ -25,6 +25,10 @@ ShufflePlanes
    Properties such as subsampling are determined from the relative size of the
    given planes to combine.
 
+   The argument *prop_src* is used to specify the output props of the clip.
+   By default it will grab the first node of *clips*. You can pass an empty array
+   to remove the props entirely from the output clip.
+
    ShufflePlanes accepts clips with variable format and dimensions only when
    extracting a single plane.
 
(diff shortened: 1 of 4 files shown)


myrsloik:
I'll have to think about this one. Why is prop_src even an array for ShufflePlanes? You only need one or zero clips as the source. No array would work just fine I think.
But this is iffy. I'd rather see a family of property manipulation functions that suck less because once you do this you'll have the next function that needs prop modification integrated.

Setsugennoao:
> I'll have to think about this one. Why is prop_src even an array for ShufflePlanes? 

You only need one or zero clips as the source. No array would work just fine I think.
Not-specified for clips[0] fallback, then empty to remove props and one node to use that.
And you can't set empty without having an array 🤔 

As for the interface I have no idea how to implement it differently without breaking scripts.

myrsloik:
Let's think this through one more round. If we add a simple (vnode:opt) to ShufflePlanes what we lose is effectively an easy way to clear properties. This can be solved by feeding a blankclip to prop_src. Still better that "not quite an array" and almost as good.

SplitPlanes is logically not even about shuffling things around so the only thing that would even conceptually make sense is an extra argument to clear all properties. 

If we add ClearProps or a similar function to clear properties (without having to go CopyProps with blankclip source) it actually becomes sane.

Summary:
- Add ClearProps()
- Add a prop_src:vnode:opt; argument to ShufflePlanes where the default is clips[0] to match the current behavior
- Maybe add something like a strip_props argument to SplitPlanes

Then we ask ourselves if we solved all the problems. For example does ClearProps need to have an argument that only makes it _ChromaLocation, _ColorRange, _Primaries, _Matrix, _Transfer. And maybe an option to only clear all properties that start with an underscore. I mean the main problem as you say is colorspace properties that become mismatched and editing them automatically creates even more weirdness (we'd need a third range option that effectively specified limited UV plane range). Thoughts?

myrsloik:
A slightly different solution implemented

PR #1019, opened 2024-01-21
Add `approximate_gamma` parameter to vsresize

saindriches:
Allow users to control it while maintaining the current behavior (bb91ba4) unchanged.
This results in a significant difference in the transfer related to HLG due to the faster implementation in zimg, which is actually another function. See:
https://github.com/sekrit-twc/zimg/blob/71431815950664f1e11b9ee4e5d4ba23d6d997f1/src/zimg/colorspace/gamma.cpp#L287-L291

For most other cases it should be okay.

### DIFF
--- a/doc/functions/video/resize.rst
+++ b/doc/functions/video/resize.rst
@@ -1,7 +1,7 @@
 Resize
 ======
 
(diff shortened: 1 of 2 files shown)


NSQY:
The Avisynth build of zimg resize has had approximate_gamma enabled by default for years. The performance difference is enormous.

PR #1044, opened 2024-03-25
fix install ps script + opinionated changes

couleurm:


myrsloik:
The script is assembled by make_portable.bat so that part is pointless. It also tries to make the script version independent which it most certainly won't be.
I've also never noticed this progress bar slowness. If you can prove it actually matters I can add a switch to hide it.

couleurm:
1. noticed just after making the pr, good call
2. progresspreference it's already specified in your script

PR #1049, opened 2024-04-09
Create a conda build recipe for conda support

sdy623:
We can build the conda packages now!


YomikoR:
There's a maintained alternative https://github.com/conda-forge/vapoursynth-feedstock

myrsloik:
Not going to be maintained by the core project. Make it available in some other way and I'll mention it as an installation method

PR #1052, opened 2024-04-25
vspipe: print messages in local encodings

random-patch-method:
I suppose the json output is never affected by local encodings. Is that correct?


myrsloik:
Correct, json file uses fprintf and will always be written as utf8.

PR #1059, opened 2024-05-09
Hide progress on powershell install script.

Zerowalker:
This improves download speed quite a bit when using Invoke-WebRequest.


myrsloik:
Personal preference. Saving one second isn't meaningful.

couleurm:
running it via pwsh 7 fixes this btw

PR #1068, opened 2024-06-12
Also suggest missing plugins in node AttributeErrors

arch1t3cht:
This is something that new users run into a lot, so it might be good to make the error message more helpful.


myrsloik:
What would be cool would be if it looks at the loaded plugin namespaces and suggests correction (if a typo). Same for functions.
And after that having a list of known namespaces (or vsrepo integration? whatever is easier) so it can say things are a plugin but not installed.

PR #1114, opened 2025-02-26
port simd to arm with sse2neon.h

yuygfgg:
Copied x86 SSE code to arm_neon folder, and translated with sse2neon.h. Also add cpu_level `neon` for arm devices (`VS_TARGET_CPU_ARM`), add ability to detect cpuFeature for arm devices.

PS. I have no idea why the commit log is like this... Please help :(


Ammar-Sadaoui:
let's hope to see android build soon

PR #1121, opened 2025-04-09
vspipe: redirect unwanted stdout to stderr

### LINKED ISSUE #1120: vspipe don't filter out script messages
I have a script that is using [vs-animesr](https://github.com/HolyWu/vs-animesr)
The problem is that I can't use vspipe to output the video stream to the encoder.
If I run the command: vspipe.exe --info "animesr_test.vpy" 
I get the following messages:

```
Warning: Unable to import quantization op. Please install modelopt library (https://github.com/NVIDIA/TensorRT-Model-Optimizer?tab=readme-ov-file#installation) to add support for compiling quantized models
Unable to import quantization op. Please install modelopt library (https://github.com/NVIDIA/TensorRT-Model-Optimizer?tab=readme-ov-file#installation) to add support for compiling quantized models
Warning: TensorRT-LLM is not installed. Please install TensorRT-LLM or set TRTLLM_PLUGINS_PATH to the directory containing libnvinfer_plugin_tensorrt_llm.so to use converters for torch.distributed ops
TensorRT-LLM is not installed. Please install TensorRT-LLM or set TRTLLM_PLUGINS_PATH to the directory containing libnvinfer_plugin_tensorrt_llm.so to use converters for torch.distributed ops
[03/30/2025-20:56:28] [TRT] [W] Functionality provided through tensorrt.plugin module is experimental.
Width: 1280
Height: 960
Frames: 855
FPS: 30000/1001 (29.970 fps)
Format Name: YUV420P10
Color Family: YUV
Alpha: No
Sample Type: Integer
Bits: 10
SubSampling W: 1
SubSampling H: 1`

The problems is that these messages are provided in input to the encoder (in this case x265), so that I get the following error message

`x265 [error]: unable to open input file <->
Warning: TensorRT-LLM is not installed. Please install TensorRT-LLM or set TRTLLM_PLUGINS_PATH to the directory containing libnvinfer_plugin_tensorrt_llm.so to use converters for torch.distributed ops
Arguments: ()
Error: fwrite() call failed when writing frame: 0, plane: 0, errno: 22
```

Workaround:

I was able to encode the same script using:

1) Vapoursynth Editor
2) NVenc -vpy-mt
3) ffmpeg -f vapoursynth -i 

all these applications have in common that they are using vapoursynth api to generate the video stream and the error message are filter out so that are not provided in input to the encoder.

I'm wondering if is possible to add an option to vspipe to achieve the same behaviour.

Thank you,
Dan

L4cache:
fixes #1120

### DIFF
--- a/src/common/wave.h
+++ b/src/common/wave.h
@@ -24,8 +24,8 @@
 #include <cstdint>
 #include <cstddef>
 
-#if !defined(UUID_T_DEFINED) && !defined(uuid_t)
-struct uuid_t {
+#if !defined(wave_uuid_t)
+struct wave_uuid_t {
     uint32_t data1;
     uint16_t data2;
     uint16_t data3;
@@ -46,7 +46,7 @@ struct WaveFormatExtensible {
     uint16_t cbSize;
     uint16_t wValidBitsPerSample;
     uint32_t dwChannelMask;
-    uuid_t  SubFormat;
+    wave_uuid_t  SubFormat;
 };
 
 struct WaveHeader {
@@ -61,13 +61,13 @@ struct WaveHeader {
 };
 
 struct Wave64Header {
-    uuid_t riffUuid;
+    wave_uuid_t riffUuid;
     uint64_t riffSize;
-    uuid_t waveUuid;
-    uuid_t fmtUuid;
+    wave_uuid_t waveUuid;
+    wave_uuid_t fmtUuid;
     uint64_t fmtSize;
     WaveFormatExtensible wfx;
-    uuid_t dataUuid;
+    wave_uuid_t dataUuid;
     uint64_t dataSize;
 };
 
--- a/src/vspipe/vspipe.cpp
+++ b/src/vspipe/vspipe.cpp
@@ -38,6 +38,8 @@
 #include <io.h>
 #include <fcntl.h>
 #include "../common/vsutf16.h"
+#else
+#include <unistd.h>
 #endif
 
 #define __STDC_FORMAT_MACROS
@@ -917,10 +919,36 @@ int main(int argc, char **argv) {
     }
 
     FILE *outFile = nullptr;
+    FILE *realStdout = nullptr;
     bool closeOutFile = false;
 
     if (opts.outputFilename.empty() || opts.outputFilename == "-") {
-        outFile = stdout;
+#ifdef VS_TARGET_OS_WINDOWS
+        realStdout = OpenFile(std::filesystem::u8path("CON"));
+        if (!realStdout) {
+            fprintf(stderr, "Failed to open new stdout\n");
+            realStdout = stdout;
+        } else if (_dup2(_fileno(stdout), _fileno(realStdout)) == -1) {
+            fprintf(stderr, "Failed to redirect stdout to new stdout\n");
+            realStdout = stdout;
+        } else if (_dup2(_fileno(stderr), _fileno(stdout)) == -1) {
+            fprintf(stderr, "Failed to redirect old stdout to stderr\n");
+            realStdout = stdout;
+        }
+#else
+        realStdout = OpenFile(std::filesystem::u8path("/dev/tty"));
+        if (!realStdout) {
+            fprintf(stderr, "Failed to open new stdout\n");
+            realStdout = stdout;
+        } else if (dup2(fileno(stdout), fileno(realStdout)) == -1) {
+            fprintf(stderr, "Failed to redirect stdout to new stdout\n");
+            realStdout = stdout;
+        } else if (dup2(fileno(stderr), fileno(stdout)) == -1) {
+            fprintf(stderr, "Failed to redirect old stdout to stderr\n");
+            realStdout = stdout;
+        }
+#endif
+        outFile = realStdout;
     } else if (opts.outputFilename == "." || opts.outputFilename == "--") {
         // do nothing
     } else {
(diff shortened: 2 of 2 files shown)


L4cache:
Can't undef typedef...

L4cache:
Any comments?

myrsloik:
Still thinking about it

dan64:
I can confirm that this patch is working as expected. In this way to the pipe are propagated only the frames (this should be also the correct design). No more conflicts with AI packages...

myrsloik:
Why are the AI packages writing to stdout anyway? Which packages are a problem?

dan64:
The problem with AI packages is that most of them are written for research purposes and/or are in development phase. An example of script having this problem is reported in this post: [#1120](https://github.com/vapoursynth/vapoursynth/issues/1120#issuecomment-2787203271).
But I found other AI packages that are able to write to stdout in a way that cannot be filtered out.
Now that are available AI packages that are able to provide useful image elaboration I think that is necessary to apply this patch for having a bullet-proof vspipe. Since vspipe is used in 99.9% of cases to provide frames to encoders via pipe, it is very important that vspipe is able to provide in output only frames by design. With this patch the stdout will be still visible but will not be propagate to the pipe. If you are unsure about the implementation of this patch, please add it at least as an option.

myrsloik:
Somewhat related, maybe named pipe support on windows would be a solution that doesn't require a horrible hack like this to be implemented.

myrsloik:
"Code is like asbestos, it's best when you don't poke it"

myrsloik:
Named pipes is what you get in R72.

L4cache:
Yes, named pipe is more ideal.

dan64:
In my knowledge named pipe is working from command line only in Linux. On windows it is missing a easy command line interface. Moreover the encoders used as frames consumer are expecting the input from a standard pipe. So it is not clear which design will be implemented and if it will be compatible with the current vspipe usage like the "horrible hack" proposed by @L4cache.

myrsloik:
> In my knowledge named pipe is working from command line only in Linux. On windows it is missing a easy command line interface. Moreover the encoders used as frames consumer are expecting the input from a standard pipe. So it is not clear which design will be implemented and if it will be compatible with the current vspipe usage like the "horrible hack" proposed by @L4cache.

Stop your FUD bullshit. The committed code is there so you can actually evaluate things yourself BEFORE writing this. But you didn't because that would mean no criticism or drama potential.

Anything that supports non-seekable file handles as input will work properly in windows.

dan64:
Sorry, I don't have intention of offending anyone and I am not using any FUD (Fear, uncertainty, and doubt) tactics. I'm just trying to understand if it is possible to have a vspipe implementation that can only provide frames via pipe by design. Unless I missed to see something I noted that only the help section of vspipe was updated. But the named pipes in windows need to be created by the code using them. There is the project [windows-named-pipe-utils](https://github.com/psmay/windows-named-pipe-utils) witch provides the necessary support programs that can be used in scripts. If the solution is this one I will try this solution, but it is still not clear to me how in this way the problem will be solved: "Idealism increases in direct proportion to one's distance from the problem." ([John Galsworthy](https://www.brainyquote.com/authors/john-galsworthy-quotes)

L4cache:
With the support written into vspipe, you don't need third party tools. And I'd say it's complete.

Mr-Z-2697:
Unfortunately some popular downstream applications don't handle named pipe very well.
For example x264 and x265, while they should be easy to support since they support pipe, they did't recognise the `\\.\pipe\` pattern as an indication to another type of pipe.

myrsloik:
> Unfortunately some popular downstream applications don't handle named pipe very well. For example x264 and x265, while they should be easy to support since they support pipe, they did't recognise the `\\.\pipe\` pattern as an indication to another type of pipe.

Go file a bug report for those projects then. Implementing support should be ~5 lines of code. Or do x264 encoding through ffmpeg.
You could also (probably) use ffmpeg to open the named pipe and then pipe it into x264 or x265 for pure insanity. Or have some other named pipe in=>stdout commandline helper.

Mr-Z-2697:
I did try to resolve that and it's less than 5 lines LOL.

x264:
```diff
diff --git a/common/osdep.h b/common/osdep.h
index 23739b7a..bb3d41d2 100644
--- a/common/osdep.h
+++ b/common/osdep.h
@@ -242,6 +242,8 @@ static inline int x264_vfprintf( FILE *stream, const char *format, va_list arg )
 static inline int x264_is_regular_file_path( const char *path )
 {
     int ret = -1;
+    if (!strncmp(path,"\\\\.\\pipe\\",9))
+        return 0;
     wchar_t path_buf[MAX_PATH];
     wchar_t *path_utf16 = x264_utf8_to_utf16_try_buf( path, path_buf, MAX_PATH );
     if( path_utf16 )
```
x265:
```diff
diff --git a/source/input/y4m.cpp b/source/input/y4m.cpp
index 186499bd1..3c42ff32e 100644
--- a/source/input/y4m.cpp
+++ b/source/input/y4m.cpp
@@ -109,7 +109,7 @@ Y4MInput::Y4MInput(InputFileInfo& info, bool alpha, int format)
     info.frameCount = -1;
     size_t estFrameSize = framesize + sizeof(header) + 1; /* assume basic FRAME\n headers */
     /* try to estimate frame count, if this is not stdin */
-    if (ifs != stdin)
+    if (ifs != stdin && strncmp(info.filename, "\\\\.\\pipe\\",9))
     {
         int64_t cur = ftello(ifs);
         if (cur >= 0)
@@ -123,7 +123,7 @@ Y4MInput::Y4MInput(InputFileInfo& info, bool alpha, int format)
     }
     if (info.skipFrames)
     {
-        if (ifs != stdin)
+        if (ifs != stdin && strncmp(info.filename, "\\\\.\\pipe\\",9))
             fseeko(ifs, (int64_t)estFrameSize * info.skipFrames, SEEK_CUR);
         else
             for (int i = 0; i < info.skipFrames; i++)
```

myrsloik:
Tell me if you actually get it merged upstream so I can write the minimum supported versions.

PR #1143, opened 2025-08-28
Align stubs, runtime, and expected behavior in Cython module

Ichunjo:
To better align the stubs, expected behavior, and runtime behavior, I propose the following changes:

* Add *abstract* base methods for `RawFrame` and `RawNode`.

* Update `FrameProps`:
  Currently, it is registered as a `Mapping`.
  The stubs annotate it as a `MutableMapping`.
  At runtime, it already behaves like a `MutableMapping`.
  I propose to formally register it as a `MutableMapping` with the following adjustments:

    * `__iter__`: make iteration independent and faster by using C objects.
    * `keys`, `items`, `values`: return the expected view objects, consistent with `MutableMapping`.
    * `pop`, `get`, `setdefault`: support positional-only parameters, as in `MutableMapping`.
    * Remove docstrings so they are not retrievable at runtime.
    * Remove redundant calls to `self.frame._ensure_open()`, since these checks are already performed in higher-level dunder methods.

* Additional cleanup commits:

  * Remove Python 2 legacy `<>` operator.
  * Update parameter names to follow Python conventions (e.g., `def __add__(self, other):` instead of `__add__(x, y):`).
  * Remove usage of deprecated legacy properties.

### DIFF
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -50,7 +50,7 @@ import keyword
 from threading import local as ThreadLocal, Lock, RLock
 from types import MappingProxyType
 from collections import namedtuple
-from collections.abc import Iterable, Mapping
+from collections.abc import ItemsView, Iterable, KeysView, MutableMapping, ValuesView
 from concurrent.futures import Future
 from fractions import Fraction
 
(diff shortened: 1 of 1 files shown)


myrsloik:
What's the point of adding the "abstract" methods to RawNode and RawFrame? They can't even be instantiated outside of cython so nobody will ever hit that code.

Ichunjo:
The goal is to align the stubs, runtime, and expected behavior in the inheritance hierarchy.

By adding abstract methods to `RawNode`/`RawFrame`, we make the intended interface explicit, so child classes are clearly required to implement them. Right now, the stubs claim these methods exist, but at runtime they don’t (`hasattr(RawNode, "get_frame") -> False`), which is misleading.

If we went stricter and only specified the methods present in `RawNode`/`VideoNode`/`AudioNode`, it would break designs like this:

```py
def func(node: RawNode) -> None:
    node.get_frame(0)  # type-check error here if get_frame is removed
```

Ichunjo:
> They can't even be instantiated outside of cython

It's probably useless on the python side but `RawNode.__new__(RawNode)` actually instantiate a new RawNode object

PR #1154, opened 2025-10-04
fix max_cache_size setter crash

jsaowji:
regression https://github.com/vapoursynth/vapoursynth/pull/1143

integer overflow probably, restore pre PR code

```py
from vapoursynth import core

a = core.max_cache_size
print(a)
core.max_cache_size = a
print(a)
```
```
4096
zsh: IOT instruction  python repro.py
```

### DIFF
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -2681,7 +2681,8 @@ cdef class Core(object):
     def max_cache_size(self, int mb):
         if mb <= 0:
             raise ValueError("Maximum cache size must be a positive number")
-        cdef int64_t new_size = mb * 1024 * 1024
+        cdef int64_t new_size = mb
+        new_size = new_size * 1024 * 1024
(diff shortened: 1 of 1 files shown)


4re:
Maybe the problem was that the original code didn't convey why that was done?
Explicit casting before doing the math would make it clear: `cdef int64_t new_size = (<int64_t>mb) * 1024 * 1024` or `cdef int64_t new_size = mb * 1024 * 1024LL`

myrsloik:
> Maybe the problem was that the original code didn't convey why that was done? Explicit casting before doing the math would make it clear: `cdef int64_t new_size = (<int64_t>mb) * 1024 * 1024` or `cdef int64_t new_size = mb * 1024 * 1024LL`

Would it? What are the mixed python and C type promotion rules?

L4cache:
Why not cast to uint64_t though, do we have negative memory?

myrsloik:
setMaxCacheSize takes an int64_t
0/negative means use defaults

PR #1216, opened 2026-06-07
Chroma location awareness in MaskedMerge and PreMultiply

### LINKED ISSUE #1157: MaskedMerge(first_plane=True) doesn't take chroma location into account
When using `MaskedMerge(first_plane=True)` to resize the mask's first plane to subsampled size for the chroma planes, it doesn't take chroma location into account and hence always results in the mask's chroma planes to be center-aligned, which will be wrong when being applied to a video with left-aligned chroma.

```py
import vapoursynth as vs
from vapoursynth import core

clip = core.bs.VideoSource(r"clip.mp4")
mask = clip.std.Sobel()

# handling in MaskedMerge(first_plane=True) which doesn't take _ChromaLocation prop into account
mask1 = mask.std.ShufflePlanes(planes=0, colorfamily=vs.GRAY)
mask1 = mask1.resize.Bilinear(width=clip.width >> clip.format.subsampling_w, height=clip.height >> clip.format.subsampling_h)
mask1 = core.std.ShufflePlanes([mask, mask1], planes=[0, 0, 0], colorfamily=vs.YUV)

# handling which take _ChromaLocation prop into account
mask2 = mask.std.ShufflePlanes(planes=[0, 0, 0], colorfamily=vs.YUV, prop_src=clip)
mask2 = mask2.resize.Bilinear(format=clip.format)

clip = core.std.Interleave([mask1, mask2])

clip.set_output()
```

https://slow.pics/c/vb0amUY3

https://github.com/user-attachments/assets/52fd2e11-89e6-4003-a723-89f55c1a0dba

JustinTArthur:
When a single-plane mask/alpha is applied to a YUV-family clip with subsampled chroma in `std.MaskedMerge` or `std.PreMultiply`, the mask/alpha now shifts to compensate for UV plane chroma location during downsample to chroma resolution. The mask now aligns to where the original luma plane sits.

It should cover cases where the frames change chroma location within the same clip and should cover cases where the clip (and the mask) are interlaced. A unit test is added to check the various scenarios using similar reference flows as  Holy Wu's code in the vapoursynth/vapoursynth#1157 description.

A prior iteration added extra resampling params to the plugin funcs. This has since been reverted.

Fixes vapoursynth/vapoursynth#1157. Also fixes a longtime crash issue when attempting to `PreMultiply` a grayscale clip (not just grayscale alpha).

### DIFF
--- a/doc/functions/video/maskedmerge.rst
+++ b/doc/functions/video/maskedmerge.rst
@@ -9,9 +9,14 @@ MaskedMerge
    The *mask* clip is assumed to  be full range for all planes and in the 
    0-1 interval for float formats regardless of the colorspace.
    If *mask* is a grayscale clip or if *first_plane* is true, the mask's first
-   plane will be used as the mask for merging all planes. The mask will be
-   bilinearly resized if necessary.
-   
+   plane will be used as the mask for merging all planes.
+
+   When a single mask plane is applied to a YUV clip with subsampled chroma,
+   it's downscaled to the chroma resolution with bilinear resampling. This
+   resampling is chroma-location aware: each frame's ``_ChromaLocation``
+   property (guessing 0, *left*, when absent) determines the sub-pixel shift
+   needed to keep the mask aligned with the luma plane.
+
    If *premultiplied* is set the blending is performed as if *clipb* has been pre-multiplied
    with alpha. In pre-multiplied mode it is an error to try to merge two frames with
    mismatched full and limited range since it will most likely cause horrible unintended
--- a/doc/functions/video/premultiply.rst
+++ b/doc/functions/video/premultiply.rst
@@ -12,3 +12,10 @@ PreMultiply
    Note that limited range pre-multiplied contents excludes the offset. For example with
    8 bit input 60 luma and 128 alpha would be calculated as ((60 - 16) * 128)/255 + 16
    and not (60 * 128)/255.
+
+   When *clip* is a YUV clip with subsampled chroma, *alpha* is resized
+   down to the chroma resolution with bilinear resampling before being
+   multiplied into the chroma planes. This resampling is chroma-location
+   aware: each frame's ``_ChromaLocation`` property (guessing 0, *left*, when
+   absent) determines the sub-pixel shift needed to keep *alpha* aligned with
+   the luma plane.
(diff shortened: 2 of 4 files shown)


myrsloik:
We're not a general resizer, I'd say we probably can assume correctly set chroma location property or simply guess it. That's all I thought of so far. Will ponder things some more but in general it looks reasonable.

myrsloik:
Two more things:
Do you ever compare the chroma location of clipa and clipb in maskedmerge to make sure the clips themselves are compatible? I don't think I saw it.

I believe the chroma location fallback in zimg is resolution dependent unlike here. No idea if we should do the same. It's just an observation and something to debate.
https://github.com/vapoursynth/vapoursynth/pull/1216/changes/ccc3da1ccd84364c5244cb4a18015f3f0818d54e#diff-845256d79e582101be26bd4b9e1917eaaea4517a071aa6fb66a8fffd093aaa49R184

JustinTArthur:
> We're not a general resizer, I'd say we probably can assume correctly set chroma location property or simply guess it. 

Up to you on this one. I like having the kernel options even if we drop `chromaloc_in[_s]`. There's not an immediately intuitive way for someone to scale a single plane mask to the 4:2:0 resolutions with their own params with `resize` alone. Holy Wu's reference example from vapoursynth/vapoursynth#1157 is close, but in its current state would only work for left-located source material without some tweaks.


> Do you ever compare the chroma location of clipa and clipb in maskedmerge to make sure the clips themselves are compatible?

You're right, I don't. It uses `clipa`'s chroma location. I like the idea of erroring if both sides have an explicit conflicting `_ChromaLocation`. If you want this, let me know and I'll add it.


> I believe the chroma location fallback in zimg is resolution dependent unlike here.

I don't think so. There's a halving that happens for interlaced, which we respect. Otherwise, we match the *guess-left* behavior employed by both `resize` and zimg internal. Are you thinking of the matrix coefficients guesser, which does have a resolution table?

myrsloik:
Ok, then we decide this:
1. We drop chromaloc_in*
2. We add th clipa and clipb matching chroma location check
Then we merge

JustinTArthur:
Force-pushed a simplified version of the changes. Adds the `clipa`/`clipb` location match check (including when it defaults to left chroma). No new `PreMultiply` or `MaskedMerge` function params.

If someone other than me wants those extra UV resampling parameters later, let me know and I'm happy to revisit.

PR #1005, opened 2023-11-21
Remove ability to destroy the Environment and Core while futures referring to it are still running

Setsugennoao:
This makes futures in get_frame_async/frames increment the reference count of the Environment so that it _can't_ be automatically destroyed by the garbage collector until they're done, as it should be since they can't be stopped.

Also, `EnvironmentPolicyAPI.destroy_environment` NEVER gets called outside of main-thread in normal operation, it should be plain and simply disallowed. A user could try doing something smart and fuck-up. This just does an extra check. @sekrit-twc also explains it in here: https://github.com/vapoursynth/vapoursynth/issues/853#issuecomment-1081269611

PR #1236, opened 2026-07-03
Wait for pending futures before environment teardown

Ichunjo:
Replace the `_get_handle_future` closure with a `HandleFuture` cdef class that tracks futures via a `WeakSet` `env.active_futures`. `destroy_environment` now calls `wait_futures` before freeing nodes, ensuring pending async frame requests resolve before invalidation.

This is somewhat similar to #1005 but:
- Setsu's PR was returning early and logged a warning. Here, `destroy_environment` is blocking. 
- `_FrameFuture` was keeping a circular reference of the EnvironmentData. Here we just track them in a WeakSet.


myrsloik:
The tests failed

Ichunjo:
Fixed. I forgot to decorate the new test by `@subprocess_runner`

PR #1253, opened 2026-08-06
Add an AArch64 NEON interpreter for std.Expr

StuartCameronCode:
`std.Expr` has no JIT backend on AArch64 — `compile_jit()` in `src/core/expr/jitcompiler.cpp` is wrapped in `#ifdef VS_TARGET_CPU_X86` — so it falls back to `ExprInterpreter::eval()`, which walks the whole bytecode program **once per pixel** through a switch, with the register file in memory. This adds a NEON interpreter over the same bytecode, evaluating four pixels at a time, keeping the scalar interpreter as the reference implementation for the row remainder.

- **Dispatch follows the genericfilters pattern**: inline in `exprfilter.cpp` under `VS_TARGET_CPU_ARM64`, gated on `VS_CPU_LEVEL_NEON` so `SetMaxCPU("none")` falls back to scalar, with the kernels in their own TU (`src/core/expr/interpreter_neon.cpp`). The header declares no NEON types, so `exprfilter.cpp` includes it unconditionally.
- **No feature gate needed.** NEON is the AArch64 baseline, and the f16 path uses only the FCVTL/FCVTN conversions, which are baseline as well — the same reasoning as the half tier in #1244.
- **Bit-identical to the scalar interpreter**, with one stated exception: the sign bit of a NaN, which is only observable on a half or single float output clip and which the x86 backends do not agree on either (the AVX2 JIT fuses multiply-add where SSE2 does not). Verified by differentially fuzzing 18000 random programs across every opcode and nine load/store format combinations.
- **Two floating-point traps worth recording.** The arithmetic opcodes are written as plain C++ operators on the vector type rather than as `vaddq_f32`/`vfmaq_f32` calls, because under `-ffp-contract=on` a compiler fuses the scalar interpreter's `SRC2 * SRC3 + SRC1` into a single multiply-add but keeps the explicit negation in `-(SRC2 * SRC3) + SRC1`; writing the same expression is what makes it reach the same decision in both, where hand-picked intrinsics cannot track it. And min/max are an explicit compare-and-select, since `std::max(a, b)` is `a < b ? b : a`, which differs from `vmaxq_f32` for NaN and for `max(-0.0, +0.0)`.
- **Two tests come with it.** The existing Expr tests build the clip with `BlankClip`'s default 640x480 and assert a single pixel, so a backend that mishandles the partial vector at the end of a row, or whose vector path disagrees with its own remainder path, passes all 65 of them. `test_expr_row_tail66` evaluates the same pixels at every offset relative to the vector boundary; `test_expr_widths67` checks every pixel for each sample format at widths either side of the usual vector widths. Both are backend-agnostic and pass on x86 unchanged.

## Benchmarking

12 expressions from the in-tree havsfunc corpus (`test/expr_compiler/havs_exprs.txt`), 720x576, median fps, comparing `SetMaxCPU("none")` against `SetMaxCPU("neon")` in the same binary.

**Apple M1** (4x Firestorm + 4x Icestorm, Apple clang 21)

| format | 1thr scalar | 1thr neon | ratio | 8thr scalar | 8thr neon | ratio |
|--------|------------:|----------:|------:|------------:|----------:|------:|
| 8-bit  | 97.3 | 483.4 | 4.97x | 326.6 | 2076.8 | 6.36x |
| 16-bit | 98.8 | 492.5 | 4.98x | 325.2 | 1962.3 | 6.03x |
| float  | 99.2 | 459.1 | 4.63x | 330.1 | 1890.6 | 5.73x |

**Rockchip RK3588** (4x Cortex-A76 @ 2.4 GHz + 4x Cortex-A55, clang 19, Ubuntu 22.04)

| format | 1thr scalar | 1thr neon | ratio | 8thr scalar | 8thr neon | ratio |
|--------|------------:|----------:|------:|------------:|----------:|------:|
| 8-bit  | 45.5 | 165.1 | 3.63x | 215.6 | 764.5 | 3.55x |
| 16-bit | 44.7 | 173.6 | 3.88x | 204.3 | 749.0 | 3.67x |
| float  | 45.6 | 163.3 | 3.58x | 206.8 | 809.8 | 3.92x |

The RK3588 1-thread figures were re-run under `taskset -c 4-7` (45.2 / 174.8 / 165.8) to confirm the scheduler had already placed the thread on an A76, since CPU 0 is an efficiency core on that part. The two 8-thread columns should not be read against each other — the core mixes differ.

The narrower win on the A76 is consistent with its two 128-bit NEON pipes against Firestorm's four, but that is an explanation rather than a measured attribution.

On real graphs this is worth a lot: QTGMC on the M1 goes 47.9 to 123.0 fps on Preset Faster, 33.1 to 74.8 on Slow and 25.6 to 49.5 on Slower, `std.Expr` having been 55% of that job. The full test suite passes on both machines, 167/167 including the two new tests.

## Note for aarch64 reviewers

Unrelated to this branch, but you hit it before reaching any of this code: at current master, building on aarch64 needs clang >= 17 or GCC >= 13, because `src/core/float16_helper.h` uses `_Float16` in C++ and `src/core/averageframesfilter.cpp` uses `__builtin_roundevenf`. Stock Ubuntu 22.04 (gcc 11.4 / clang 15) cannot build the tree at all. Flagging it so it is not misattributed to this change.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -445,6 +445,7 @@ else
     if enable_arm_asm
         # NEON is the AArch64 baseline; its kernels ride with the plugin proper.
         arm_filter_sources += files(
+            'src/core/expr/interpreter_neon.cpp',
             'src/core/kernel/arm/convolution_neon.cpp',
             'src/core/kernel/arm/generic_neon.cpp',
             'src/core/kernel/arm/lut_neon.cpp',
--- a/src/core/expr/interpreter_neon.cpp
+++ b/src/core/expr/interpreter_neon.cpp
@@ -1,0 +1,229 @@
+/*
+* Copyright (c) 2012-2020 Fredrik Mellbin
+*
+* This file is part of VapourSynth.
+*
+* VapourSynth is free software; you can redistribute it and/or
+* modify it under the terms of the GNU Lesser General Public
+* License as published by the Free Software Foundation; either
+* version 2.1 of the License, or (at your option) any later version.
+*
+* VapourSynth is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+* Lesser General Public License for more details.
+*
+* You should have received a copy of the GNU Lesser General Public
+* License along with VapourSynth; if not, write to the Free Software
+* Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+*/
+
+// NEON evaluator for Expr bytecode. compile_jit() only has an x86 backend, so
+// without this every std.Expr on aarch64 runs ExprInterpreter::eval() once per
+// pixel: a switch over the whole program per pixel, with the register file in
+// memory. Evaluating each opcode across a vector instead amortizes the dispatch
+// and does the arithmetic four-wide.
+//
+// Every opcode below mirrors ExprInterpreter::eval() exactly, so the two produce
+// identical results. Two things that mirroring depends on, both easy to get
+// wrong:
+//
+//   - The arithmetic opcodes are written as plain C++ operators on the vector
+//     type, not as vaddq_f32/vmulq_f32/vfmaq_f32 calls. With the default
+//     -ffp-contract=on a compiler fuses the scalar interpreter's
+//     `SRC2 * SRC3 + SRC1` into one multiply-add (rounding once) but keeps the
+//     explicit negation in `-(SRC2 * SRC3) + SRC1` (rounding twice, and flipping
+//     the sign bit of a NaN product). Writing the same expression is what makes
+//     the compiler reach the same decision in both; hand-picked intrinsics
+//     cannot track it.
+//   - std::max(a, b) is `a < b ? b : a` and std::min(a, b) is `b < a ? b : a`,
+//     which differ from vmaxq_f32/vminq_f32 for NaN and for max(-0.0, +0.0), so
+//     they are written as an explicit compare and select.
+//
+// The transcendentals have no NEON equivalent and are evaluated per lane through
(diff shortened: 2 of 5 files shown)


myrsloik:
This one is trickier to justify. This plugin exists with a proper JIT: https://github.com/Jaded-Encoding-Thaumaturgy/akarin-vapoursynth-plugin/releases

I've been thinking about how to rework Expr in general. Soon GPU Expr will be a thing as well. Consider this one parked at the moment.

StuartCameronCode:
I've tried incorporating the akarin JIT as a patch on top of VapourSynth in the latest [VapourBox](https://github.com/StuartCameronCode/VapourBox) release and it makes a massive difference to QTGMC deinterlacing speeds on arm64 - approaching x64 performance. It was a pretty easy addition for the sake of the users complaining about processing speeds until you sort out a solution up-stream. Thanks for the tip! :)

emotion3459:
> it makes a massive difference to QTGMC deinterlacing speeds on arm64 - approaching x64 performance.

The most up to date [QTGMC](https://jaded-encoding-thaumaturgy.github.io/vs-jetpack/api/vsdeinterlace/qtgmc/) implementation doesn't use std.Expr to begin with, so the real-world performance benefit should be zero.

JustinTArthur:
FWIW, this is how I would have approached it if wanting the LLVM-free path. While not as fast as a JIT might be, it's friendly to SELinux and the macOS hardened runtime. If we don't have a fancier Expr in several months, definitely put something like this back up for consideration.

JustinTArthur:
Also worth mentioning that the biggest arm64 bottleneck in old and new QTGMC is NNEDI3. There are Vulkan harnesses for it that can run on MoltenVK and [an un-reviewed znedi3 PR for significant arm64 gains](https://github.com/sekrit-twc/znedi3/pull/26).

PR #1257, opened 2026-08-22
Add support for Python free-threading

Ichunjo:
- Configured Meson and headers to support free-threaded builds (`Py_GIL_DISABLED`) without limited API (abi3t isn't supported yet), while keeping abi3 for standard Python. Bumped Cython requirement to >= 3.3.0.
- Added a `SynchronizedDict` helper (`cython.pymutex`) and locked access to environment maps, known nodes, and async frame reordering queues.
- Initialized `_vsapi` at import time to prevent lazy-init race conditions.
- Replaced manual counter / `WeakValueDictionary` with `weakref.WeakSet`.
- Added a `collect_garbage` parameter in `RawNode.frames()`. Allows to opt out of the stop-the-world `gc.collect()` when the generator exits.
- Moved `cibuildwheel` settings and wheel tests from the GitHub workflow into `pyproject.toml` (makes running locally much easier).
- Explicitly linked `threads` dependency in Meson for `libvsscript` and `vspipe` (fixes glibc < 2.34 linking).
- Updated GitHub Actions to latest versions.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -153,9 +153,10 @@ int main() {
     endif
 endif
 
+threads_dep = dependency('threads')
 deps = [
     dl_dep,
-    dependency('threads'),
+    threads_dep,
     dependency('zimg', version: '>=3.0.5'),
 ]
 
@@ -266,7 +267,7 @@ py.install_sources(
 libvsscript = library('vsscript',
     files('src/vsscript/vsscript.cpp'),
     cpp_args: ['-DPy_NO_LINK_LIB', '-DVS_CORE_EXPORTS'] + (is_freethreaded ? ['-DPy_GIL_DISABLED=1'] : []),
-    dependencies: [dl_dep, py_dep],
+    dependencies: [dl_dep, py_dep, threads_dep],
     gnu_symbol_visibility: 'hidden',
     include_directories: incdir,
     install: true,
--- a/meson.build
+++ b/meson.build
@@ -237,12 +237,16 @@ vapoursynth_dep = declare_dependency(
     link_with: libvapoursynth,
 )
 
+is_freethreaded = py.get_variable('Py_GIL_DISABLED', 0) == 1 or py.get_variable('ABIFLAGS', '').contains('t')
+py_limited_api = is_freethreaded ? '' : '3.12'
+
 py.extension_module('vapoursynth',
     files('src/common/wave.cpp', 'src/cython/vapoursynth.pyx'),
+    cython_args: ['-X', 'freethreading_compatible=True'],
     dependencies: [py_dep, vapoursynth_dep],
     install: true,
     install_rpath: '$ORIGIN',
-    limited_api: '3.12',
+    limited_api: py_limited_api,
     override_options: 'cython_language=cpp',
     subdir: 'vapoursynth',
 )
@@ -261,7 +265,7 @@ py.install_sources(
 
 libvsscript = library('vsscript',
     files('src/vsscript/vsscript.cpp'),
-    cpp_args: ['-DPy_NO_LINK_LIB', '-DVS_CORE_EXPORTS'],
+    cpp_args: ['-DPy_NO_LINK_LIB', '-DVS_CORE_EXPORTS'] + (is_freethreaded ? ['-DPy_GIL_DISABLED=1'] : []),
     dependencies: [dl_dep, py_dep],
     gnu_symbol_visibility: 'hidden',
     include_directories: incdir,
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,6 +60,13 @@ exclude = [
 build-frontend = "uv"
 build = ["cp314-win_amd64", "cp312-*", "cp315t-*"]
 skip = ["cp312-win*", "*-win32"]
+test-sources = ["test"]
+test-command = [
+    "vapoursynth config",
+    "vapoursynth register-install",
+    "vspipe --version",
+    'python -m unittest discover -s test -p "*test.py"',
+]
 
 [tool.cibuildwheel.windows]
 config-settings = { setup-args = ["--debug", "--vsenv"] }
@@ -68,6 +75,9 @@ repair-wheel-command = [
     "uvx wheel tags --remove --python-tag cp312 {wheel}",
     'move /y "{wheel}\..\*.whl" "{dest_dir}"',
 ]
+
+[tool.cibuildwheel.linux]
+test-command = 'python -m unittest discover -s test -p "*test.py"'
 
 [tool.cibuildwheel.macos]
 environment = { MACOSX_DEPLOYMENT_TARGET = "13.3" }
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["Cython>=3.1.0", "meson-python"]
+requires = ["Cython>=3.3.0", "meson-python"]
 build-backend = "mesonpy"
 
 [project]
@@ -9,9 +9,7 @@ readme = "README.md"
 requires-python = ">=3.12"
 license = "LGPL-2.1-or-later"
 license-files = ["COPYING.LESSER"]
-authors = [
-    {name = "Fredrik Mellbin", email = "fredrik.mellbin@gmail.com"},
-]
+authors = [{ name = "Fredrik Mellbin", email = "fredrik.mellbin@gmail.com" }]
 keywords = ["frameserver", "video", "audio"]
 classifiers = [
     "Development Status :: 5 - Production/Stable",
@@ -57,6 +55,31 @@ exclude = [
     "*/vspipe.pdb",
     "*/vsvfw.pdb",
 ]
+
+[tool.cibuildwheel]
+build-frontend = "uv"
+build = ["cp314-win_amd64", "cp312-*", "cp315t-*"]
+skip = ["cp312-win*", "*-win32"]
+
+[tool.cibuildwheel.windows]
+config-settings = { setup-args = ["--debug", "--vsenv"] }
+environment = { CC = "clang-cl", CXX = "clang-cl", AR = "llvm-lib" }
+repair-wheel-command = [
+    "uvx wheel tags --remove --python-tag cp312 {wheel}",
+    'move /y "{wheel}\..\*.whl" "{dest_dir}"',
+]
+
+[tool.cibuildwheel.macos]
+environment = { MACOSX_DEPLOYMENT_TARGET = "13.3" }
+
+[[tool.cibuildwheel.overrides]]
+select = "cp3??t-*"
+config-settings = { setup-args = ["-Dpython.allow_limited_api=false"] }
+
+[[tool.cibuildwheel.overrides]]
+select = "cp3??t-win*"
+config-settings = { setup-args = ["--debug", "--vsenv", "-Dpython.allow_limited_api=false"] }
+repair-wheel-command = ""
 
 [tool.ruff]
 extend-exclude = ["vapoursynth.pyi"]
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -22,8 +22,7 @@ include 'vsconstants.pxd'
 from vshelper cimport bitblt
 from vsscript_internal cimport VSScript
 from wave cimport WaveHeader, Wave64Header, CreateWave64Header, CreateWaveHeader, PackChannels16to16le, PackChannels32to24le, PackChannels32to32le
-cimport cython.parallel
-from cython cimport view, final
+cimport cython
 from libc.stdlib cimport malloc, free, realloc
 from libc.stdint cimport intptr_t, int16_t, uint16_t, int32_t, uint32_t, uint8_t, uint64_t, int64_t
 from cpython.buffer cimport PyBUF_READ, PyBUF_SIMPLE, PyBuffer_FillInfo, PyBuffer_Release
@@ -34,7 +33,6 @@ from cpython.ref cimport Py_INCREF, Py_DECREF
 import os
 import enum
 import ctypes
-import threading
 import traceback
 import gc
 import sys
@@ -229,7 +227,7 @@ class VapourSynthAPIVersion(typing.NamedTuple):
 __version__ = VapourSynthVersion(VS_CURRENT_RELEASE, 0)
 __api_version__ = VapourSynthAPIVersion(VAPOURSYNTH_API_MAJOR, VAPOURSYNTH_API_MINOR)
 
-@final
+@cython.final
 cdef class EnvironmentData(object):
     cdef bint alive
     cdef Core core
@@ -308,7 +306,7 @@ class EnvironmentPolicy(object):
         return env.alive
 
 
-@final
+@cython.final
 cdef class StandaloneEnvironmentPolicy:
     cdef EnvironmentData _environment
     cdef object _api
@@ -391,7 +389,8 @@ cdef void __stdcall _logFree(void* userData) noexcept nogil:
     with gil:
         Py_DECREF(<object>userData)
 
-@final
+
+@cython.final
 cdef class EnvironmentPolicyAPI:
     # This must be a weak-ref to prevent a cyclic dependency that happens if the API
     # is stored within an EnvironmentPolicy-instance.
@@ -622,7 +621,7 @@ def unregister_on_destroy(callback):
     env.on_destroy.remove(callback)
 
 
-@final
+@cython.final
 cdef class _FastManager(object):
     cdef EnvironmentData target
     cdef EnvironmentData previous
@@ -3668,7 +3667,7 @@ cdef void __stdcall publicFunction(const VSMap *inm, VSMap *outm, void *userData
                 vsapi.mapSetError(outm, emsg)
 
 
-@final
+@cython.final
 cdef class VSScriptEnvironmentPolicy:
     cdef dict _env_map
 
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -2287,7 +2287,7 @@ cdef class RawNode(object):
             self.funcs.getFrameAsync(n, self.node, frameDoneCallback, <void *>data)
 
 
-    def frames(self, prefetch=None, backlog=None, close=False):
+    def frames(self, prefetch=None, backlog=None, close=False, collect_garbage=True):
         if prefetch is None or prefetch <= 0:
             prefetch = self.core.num_threads
         if backlog is None or backlog < 0:
@@ -2418,7 +2418,8 @@ cdef class RawNode(object):
                 with contextlib.suppress(BaseException):
                     fut.result().close()
 
-            gc.collect()
+            if collect_garbage:
+                gc.collect()
 
     def clear_cache(self):
         self.ensure_valid()
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -352,7 +352,7 @@ cdef class StandaloneEnvironmentPolicy:
 # Internal holder of the current policy.
 cdef object _policy = None
 
-cdef const VSAPI *_vsapi = NULL
+cdef const VSAPI *_vsapi = getVapourSynthAPI(VAPOURSYNTH_API_VERSION)
 
 
 cdef void _set_logger(EnvironmentData env, VSLogHandler handler, VSLogHandlerFree free, void *userData):
@@ -4038,10 +4038,7 @@ cdef public api VSCore *vpy4_getCore(VSScript *se) nogil:
 cdef public api const VSAPI *vpy4_getVSAPI(int version) nogil:
     return getVapourSynthAPI(version)
 
-cdef const VSAPI *getVSAPIInternal() nogil:
-    global _vsapi
-    if _vsapi == NULL:
-        _vsapi = getVapourSynthAPI(VAPOURSYNTH_API_VERSION)
+cdef inline const VSAPI *getVSAPIInternal() nogil:
     return _vsapi
 
 cdef public api int vpy4_getVariable(VSScript *se, const char *name, VSMap *dst) nogil:
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -45,9 +45,9 @@ import functools
 import typing
 import warnings
 import keyword
-from threading import local as ThreadLocal, Lock, RLock
+from threading import local as ThreadLocal, Condition, Event, Lock
 from types import MappingProxyType
-from collections.abc import ItemsView, Iterable, KeysView, MutableMapping, ValuesView
+from collections.abc import ItemsView, Iterable, KeysView, Mapping, MutableMapping, ValuesView
 from concurrent.futures import Future, CancelledError as FutureCancelledError, wait as wait_futures
 from fractions import Fraction
 
@@ -227,15 +227,105 @@ class VapourSynthAPIVersion(typing.NamedTuple):
 __version__ = VapourSynthVersion(VS_CURRENT_RELEASE, 0)
 __api_version__ = VapourSynthAPIVersion(VAPOURSYNTH_API_MAJOR, VAPOURSYNTH_API_MINOR)
 
+
+cdef object _SYNCHRONIZED_SENTINEL
+try:
+    _SYNCHRONIZED_SENTINEL = sentinel("_SYNCHRONIZED_SENTINEL")
+except NameError:
+    _SYNCHRONIZED_SENTINEL = object()
+
+
+@cython.final
+@cython.internal
+cdef class SynchronizedDict:
+    cdef dict data
+    cdef cython.pymutex lock
+
+    def __cinit__(self):
+        self.data = {}
+
+    def __getitem__(self, object key):
+        with self.lock:
+            return self.data[key]
+
+    def __setitem__(self, object key, object value):
+        with self.lock:
+            self.data[key] = value
+
+    def __delitem__(self, object key):
+        with self.lock:
+            del self.data[key]
+
+    def __contains__(self, object key):
+        with self.lock:
+            return key in self.data
+
+    def __len__(self):
+        with self.lock:
+            return len(self.data)
+
+    def __iter__(self):
+        with self.lock:
+            return iter(list(self.data))
+
+    def get(self, object key, object default=None):
+        with self.lock:
+            return self.data.get(key, default)
+
+    def pop(self, object key, object default=_SYNCHRONIZED_SENTINEL):
+        with self.lock:
+            if default is _SYNCHRONIZED_SENTINEL:
+                return self.data.pop(key)
+            return self.data.pop(key, default)
+
+    def clear(self):
+        with self.lock:
+            self.data.clear()
+
+    def keys(self):
+        with self.lock:
+            return list(self.data.keys())
+
+    def values(self):
+        with self.lock:
+            return list(self.data.values())
+
+    def items(self):
+        with self.lock:
+            return list(self.data.items())
+
+    def copy(self):
+        with self.lock:
+            return dict(self.data)
+
+
+class SynchronizedDictView(Mapping):
+    def __init__(self, target):
+        self._target = target
+
+    def __getitem__(self, key):
+        return self._target[key]
+
+    def __iter__(self):
+        return iter(self._target)
+
+    def __len__(self):
+        return len(self._target)
+
+    def __repr__(self):
+        return f"{type(self).__name__}({dict(self._target.items())!r})"
+
+
 @cython.final
 cdef class EnvironmentData(object):
     cdef bint alive
+    cdef bint destroying
     cdef Core core
     cdef object on_destroy
-    cdef dict outputs
-    cdef dict active_exceptions
+    cdef SynchronizedDict outputs
+    cdef SynchronizedDict active_exceptions
     cdef int next_exc_id
-    cdef object exc_lock
+    cdef cython.pymutex lock
 
     cdef int coreCreationFlags
     cdef VSLogHandle* log
@@ -262,24 +352,28 @@ cdef class EnvironmentData(object):
         _unset_logger(self)
 
     cdef int store_exception(self, object e):
-        if self.active_exceptions is None:
+        cdef SynchronizedDict exc = self.active_exceptions
+
+        if exc is None:
             return 0
 
-        with self.exc_lock:
+        with exc.lock:
             self.next_exc_id += 1
-            self.active_exceptions[self.next_exc_id] = e
+            exc.data[self.next_exc_id] = e
 
-            if len(self.active_exceptions) > 1000:
-                self.active_exceptions.pop(next(iter(self.active_exceptions)), None)
+            if len(exc.data) > 1000:
+                exc.data.pop(next(iter(exc.data)), None)
 
             return self.next_exc_id
 
     cdef object retrieve_exception(self, str msg_str):
-        if self.active_exceptions is not None and msg_str.startswith("[VS_PY_EXC_ID:"):
+        cdef SynchronizedDict exc = self.active_exceptions
+
+        if exc is not None and msg_str.startswith("[VS_PY_EXC_ID:"):
             try:
                 parts = msg_str.split("]\n", 1)
                 exc_id = int(parts[0][len("[VS_PY_EXC_ID:"):])
-                return self.active_exceptions.pop(exc_id, None)
+                return exc.pop(exc_id, None)
             except (ValueError, KeyError, IndexError):
                 pass
         return None
@@ -349,8 +443,10 @@ cdef class StandaloneEnvironmentPolicy:
         return environment is self._environment
 
 
-# Internal holder of the current policy.
+# Internal holder of the current policy and lock.
 cdef object _policy = None
+cdef cython.pymutex _policy_lock
+_policy_ready = Event()
 
 cdef const VSAPI *_vsapi = getVapourSynthAPI(VAPOURSYNTH_API_VERSION)
 
@@ -396,7 +492,7 @@ cdef class EnvironmentPolicyAPI:
     # is stored within an EnvironmentPolicy-instance.
     cdef object _target_policy
 
-    cdef object _lock
+    cdef cython.pymutex _lock
     cdef object _known_environments
     # Sadly, weakref has no WeakSet.
     # So we use a counter to fake a WeakSet.
@@ -421,16 +517,16 @@ cdef class EnvironmentPolicyAPI:
         cdef EnvironmentData env = EnvironmentData.__new__(EnvironmentData)
         env.core = None
         env.log = NULL
-        env.outputs = {}
-        env.active_exceptions = {}
+        env.outputs = SynchronizedDict()
+        env.active_exceptions = SynchronizedDict()
         env.next_exc_id = 0
-        env.exc_lock = Lock()
         env.coreCreationFlags = flags
         env.on_destroy = []
         env.env_locals = weakref.WeakKeyDictionary()
         env.known_nodes = weakref.WeakSet()
         env.known_frames = weakref.WeakSet()
         env.active_futures = weakref.WeakSet()
+        env.destroying = False
         env.alive = True
 
         with self._lock:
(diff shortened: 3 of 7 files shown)


meson.build
```diff
@@ -272,6 +277,7 @@ libvsscript = library('vsscript',
 vsscript_dep = declare_dependency(
     include_directories: incdir,
     link_with: libvsscript,
+    dependencies: threads_dep,
```
HolyWu: Redundant. `link_with: libvsscript` already implicitly carries the dependencies on which the linked libraries depend.

meson.build
```diff
@@ -291,7 +297,7 @@ executable('vspipe',
         'src/vspipe/vsjson.cpp',
         'src/vspipe/vspipe.cpp',
     ),
-    dependencies: vsscript_dep,
+    dependencies: [vsscript_dep, threads_dep],
```
HolyWu: Same above.

Ichunjo:
Addressed the review comments

myrsloik:
Will merge after GPU support is out

src/cython/vapoursynth.pyx
```diff
@@ -455,46 +521,54 @@ cdef class EnvironmentPolicyAPI:
...
-                except Exception as e:
-                    formatted = ''.join(traceback.format_exception(type(e), e, e.__traceback__))
-                    if env.core is not None:
-                        env.core.log_message(MessageType.MESSAGE_TYPE_CRITICAL, formatted)
-                    else:
-                        sys.stderr.write(formatted)
+        with env.lock:
+            if not env.alive:
+                return
+            env.alive = False
```
HolyWu: Moving `env.alive = False` to the top of `destroy_environment` changes which environment the `on_destroy` callbacks run in.

`VSScriptEnvironmentPolicy.set_environment()` does `if not self.is_alive(environment): environment = None`, and `is_alive()` returns `environment.alive`. Previously `alive` was only cleared at the very end of this function, so the `with use_environment(env).use():` below actually activated the environment. Now `alive` is already `False` when `_FastManager.__enter__` calls `set_environment(env)`, so the thread-local stack is set to `None` and the callbacks run with **no** environment active.

_(edited to correct scope and symptom after re-checking)_

**Scope:** this affects policies whose `set_environment` filters on `is_alive` — `VSScriptEnvironmentPolicy` (vspipe/vsscript/editor hosts) and third-party policies copying that pattern. `StandaloneEnvironmentPolicy.set_environment` ignores its argument and `get_current_environment` keeps returning `self._environment` until `on_policy_cleared` runs *after* `destroy_environment` returns, so plain `import vapoursynth` usage is unaffected.

**Symptom:** the callback exceptions do **not** propagate — `destroy_environment` catches them and writes the traceback via `env.core.log_message(MESSAGE_TYPE_CRITICAL, ...)` or `sys.stderr`. So a `vs.register_on_destroy(cleanup)` whose `cleanup` touches `vs.core`, `vs.clear_outputs()` or a `vs.Local` silently stops doing its work while a CRITICAL traceback appears in the host log. Easier to miss in the field than I first implied, but not a crash.

**On the fix:** the early clear is not gratuitous — it is what makes the `if not env.alive: return` guard an atomic under-lock idempotency check. So "move it back to the end" is not the answer; it needs a separate `destroying` flag consumed by that guard with `alive` staying True through the callback phase, or `destroy_environment` setting the thread's current environment directly instead of routing through the policy's filtering `set_environment`.

Note also that no existing test covers this: `StubPolicy.set_environment` in `test/environment_test.py` is a bare `self._current = environment` with no `is_alive` filter, and the callbacks in `test_environment_destruction_runs_callbacks` only flip a list element. A regression test would need a filtering policy plus a callback that reads `vs.core`.

Related, from the same change: because the early return at the top now fires as soon as `alive` is cleared, a second `destroy_environment(env)` on another thread returns *immediately* — reporting the environment as destroyed — while the first thread is still inside `wait_futures()`, the callbacks, the node/frame sweeps and `freeCore()`. A host that treats that return as "teardown complete" and frees its own core can double-free.

src/cython/vapoursynth.pyx
```diff
@@ -267,12 +337,12 @@ cdef class EnvironmentData(object):
         if self.active_exceptions is None:
             return 0
 
-        with self.exc_lock:
+        with self.active_exceptions.lock:
```
HolyWu: `self.active_exceptions.lock` and `self.active_exceptions.data` are **C-level** attribute accesses on a `cdef SynchronizedDict`. With `nonecheck` off (the default, and no directive is set in this file) Cython emits no `None` guard, so if `active_exceptions` is `None` here this dereferences `Py_None` as a `SynchronizedDict *` and locks a garbage `PyMutex` — a hard crash, not an exception.

The `if self.active_exceptions is None: return 0` guard three lines up does not close the window: `destroy_environment` sets `env.active_exceptions = None` without taking any lock that this function observes, so a core worker thread whose filter raises during teardown can lose that race.

Before this PR the same race hit `cdef object exc_lock` / `cdef dict active_exceptions`, which are Python-object accesses that raise a catchable `AttributeError`/`TypeError`. This is a regression from "spurious exception" to "segfault".

Simplest fix: keep a local strong reference and re-check it, e.g.

```cython
cdef SynchronizedDict exc = self.active_exceptions
if exc is None:
    return 0
with exc.lock:
    ...
```

(and have `destroy_environment` hold `env.lock` while it nulls the field).

src/cython/vapoursynth.pyx
```diff
@@ -507,22 +581,29 @@ cdef class EnvironmentPolicyAPI:
...
-        env.known_nodes = None
-        env.known_frames.clear()
-        env.known_frames = None
-        env.alive = False
+        if env.env_locals is not None:
+            env.env_locals.clear()
+            env.env_locals = None
+        if env.active_exceptions is not None:
+            env.active_exceptions.clear()
+            env.active_exceptions = None
```
HolyWu: This whole teardown block — plus the `known_nodes` / `known_frames` invalidation sweeps above — runs **without** `env.lock`, while the readers this PR changed (`createVideoNode`, `createVideoFrame`, `createAudioNode`, `createConstVideoFrame`, ...) now do take `env.lock` before their `is not None` check + `add()`.

Locking only one side buys nothing. Two concrete outcomes with the GIL disabled:

1. Thread B holds `env.lock`, passes `if env.known_nodes is not None`, thread A executes `env.known_nodes = None` here → `AttributeError: 'NoneType' object has no attribute 'add'` inside node construction.
2. Thread B adds a node just after A's invalidation loop finished. That node is never invalidated, keeps its `VSNode*`, and `freeCore()` a few lines up frees the core underneath it → use-after-free on the next `get_frame`.

`destroy_environment` should hold `env.lock` for the field-nulling (and ideally publish a `destroying` flag under the lock that the create* helpers check) so the reader-side locking actually excludes it.

src/cython/vapoursynth.pyx
```diff
@@ -536,44 +617,52 @@ cdef class EnvironmentPolicyAPI:
             return f"<EnvironmentPolicyAPI bound to {target!r}>"
 
 
-def register_policy(policy):
+# This function must be invoked with the _policy_lock acquired
+cdef _register_policy_unlocked(object policy):
     global _policy
-    if _policy is not None:
-        raise RuntimeError("There is already a policy registered.")
     _policy = policy
```
HolyWu, line 652: _(edited after re-checking — the race is real but my original framing and fix were wrong)_

The defect is not the write order here; it is that `get_policy()` is the **broken variant of double-checked locking** — its fast path reads `_policy` and returns it with no lock held:

```cython
if _policy is None:
    with _policy_lock:
        if _policy is None:
            ...
            _register_policy_unlocked(standalone_policy)
return _policy
```

`_register_policy_unlocked` sets `_policy = policy` before `policy.on_policy_registered(_api)`, and for `StandaloneEnvironmentPolicy` that callback is what assigns `self._environment` — via `api.create_environment()` and `api.set_logger()`, the latter reaching `vsscript_get_core_internal` → `createCore()` → `funcs.createCore()`, which scans and loads every plugin from disk. That window is milliseconds. A second thread entering `get_policy()` in it skips the lock, returns the half-built policy, gets `None` from `get_current_environment()`, and raises.

**Do not move the assignment.** The ordering is forced: `EnvironmentPolicyAPI.ensure_policy_matches()` compares the global `_policy` against `self._target_policy()`, `create_environment()` calls it first, and `on_policy_registered` calls `create_environment`. Publishing `_policy` later makes every registration die with `ValueError("The currently activated policy does not match the bound policy...")`.

The fix has to be reader-side: add a `cdef bint _policy_ready` set only after `on_policy_registered` returns (and cleared in `clear_policy`), and have `get_policy()`'s fast path test that instead of `_policy is None`. Taking `_policy_lock` unconditionally also works here — I checked that `createCore`/`vsscript_get_core_internal` never re-enter `get_policy()` — but `cython.pymutex` is **not** reentrant, so that variant breaks the moment any policy's `on_policy_registered` calls back into `get_policy()`.

Two corrections to what I wrote above:

- The symptom quote was misattributed. `set_output` raises `Error('Internal environment id not set. set_output called from a filter callback?')`; `vs.core` goes through `_get_core()` and raises `Error('No environment is currently activated. Please activate an environment. ...')`.
- **Provenance and severity.** This is not a regression — master has the same write order in `register_policy` and its `get_policy()` had *no* lock at all (two racing threads could both reach `register_policy`, and the loser got a spurious `RuntimeError("There is already a policy registered.")`). This PR closes the duplicate-construction half and leaves the publish-before-init half open. Severity is low/medium: startup-only, needs two threads' *first* VapourSynth call to overlap with no policy pre-registered, and the failure is a clean Python exception, not corruption.

Separately, and worth its own look rather than folding in here: `return _policy` / `has_policy()` / `ensure_policy_matches` all do a borrowed read + INCREF of the module-level `cdef object _policy` with no lock, while `clear_policy` nulls it. There is no tearing (pointer-width store), but there is a use-after-free window once `clear_policy`'s caller drops the returned strong reference — `unregister_policy` discards it.

pyproject.toml
```diff
@@ -58,6 +56,41 @@ exclude = [
     "*/vsvfw.pdb",
 ]
 
+[tool.cibuildwheel]
+build-frontend = "uv"
+build = ["cp314-win_amd64", "cp312-*", "cp315t-*"]
```
HolyWu, line 61: **Retracted — this comment was wrong. Please disregard it, and in particular do not apply the change it suggested.**

I claimed free-threaded and pre-release identifiers were gated behind `enable`, and suggested adding `enable = ["cpython-prerelease", "cpython-freethreading"]`. Both halves are false for the version this PR pins, and the suggestion would have broken the build.

From `cibuildwheel/selector.py` at the pinned `v4.2.0` tag:

```python
class EnableGroup(StrEnum):
    CPythonPrerelease = "cpython-prerelease"
    GraalPy = "graalpy"
    PyPy = "pypy"
    PyPyEoL = "pypy-eol"
    PyodideEoL = "pyodide-eol"
    PyodidePrerelease = "pyodide-prerelease"
```

There is no free-threading group at all, and `parse_option_value` raises `Unknown enable group: ...` for unrecognised values — so `enable = ["cpython-freethreading"]` would have aborted cibuildwheel with exit 2. `BuildSelector.__call__` contains no filter on the `t` suffix anywhere, and the only prerelease gate is hard-coded:

```python
if EnableGroup.CPythonPrerelease not in self.enable and fnmatch(build_id, "cp316*"):
    return False
```

`resources/build-platforms.toml` at v4.2.0 also carries `cp315t-*` for every platform this workflow builds (`cp315t-win_amd64`, `cp315t-manylinux_x86_64`, `cp315t-manylinux_aarch64`, `cp315t-macosx_*`). So free-threaded builds are opt-**out** on 4.2.0, `cp315t-*` selects normally, and both `[[tool.cibuildwheel.overrides]]` blocks are live.

Sorry for the noise. The config is correct as written.

src/cython/vapoursynth.pyx
```diff
@@ -455,46 +521,54 @@ cdef class EnvironmentPolicyAPI:
...
-                        env.core.log_message(MessageType.MESSAGE_TYPE_CRITICAL, formatted)
-                    else:
-                        sys.stderr.write(formatted)
+        with env.lock:
+            if not env.alive:
+                return
+            env.alive = False
+
+            futs = env.active_futures
+            env.active_futures = None
```
HolyWu, line 555: _(edited — the structural claim holds, but I have withdrawn the WeakSet half and fixed a wrong line attribution)_

`env.active_futures` is nulled here under `env.lock`, but none of its three producers were updated — they still do an unlocked read-check-read-use:

- `CallbackData.__init__` (reached from `RawNode.get_frame_async` via `createCallbackData`) — **not** `HandleFuture`, whose `__cinit__` deliberately touches no environment; I had that wrong.
- `VideoNode.get_frame`
- `AudioNode.get_frame`

Thread A evaluates `env.active_futures is not None` → True; thread B enters `destroy_environment`, takes `env.lock`, executes `env.active_futures = None`; thread A then calls `.add(fut)` on `None` → `AttributeError: 'NoneType' object has no attribute 'add'` escapes `get_frame`/`get_frame_async`, where the guard's evident intent is a silent skip.

**Withdrawn:** my claim that `WeakSet.add()` is non-atomic and can be corrupted by concurrent adds. That is the 3.13 implementation; on this PR's free-threaded target `_IterationGuard`/`_pending_removals`/`_commit_removals` are gone from `WeakSet` and `add` is a single `set.add`, which free-threaded CPython protects with the set's own per-object lock. Same for `WeakKeyDictionary.setdefault`. No corruption. There is no use-after-free either — the local `futs` pins the WeakSet across the whole teardown.

**The more serious consequence is one I understated.** If thread A wins the `.add()` *after* thread B has snapshotted `futs`, then `wait_futures(futs)` returns without waiting for A's frame request. B proceeds to `rn.funcs.freeNode(rn.node)` over every known node while A is still inside `with nogil: self.funcs.getFrame(...)` — use-after-free of the `VSNode`. Tracking futures is precisely what is supposed to prevent that, and the unlocked add is what defeats it.

**Framing correction:** this is not "exactly the TOCTOU the PR fixed for known_nodes/known_frames". The PR locked the *producer* side there while `destroy_environment` nulls `known_nodes`/`known_frames`/`env_locals` outside `env.lock` (see my other comment). The two paths are mirror-image half-fixes: `active_futures` locks the writer but not the readers; `known_nodes`/`known_frames` lock the readers but not the writer. Worth treating as one "the `env.lock` discipline over `EnvironmentData`'s teardown-nulled containers is applied inconsistently" issue.

**Suggested fix** — snapshot into a local rather than adding another lock:

```cython
futs = env.active_futures
if futs is not None:
    futs.add(fut)
```

That removes the `AttributeError` without calling Python code while holding a `cython.pymutex`, which the `known_nodes` sites now do and which the Cython free-threading docs flag as a deadlock risk (Cython can re-acquire the GIL implicitly, e.g. when raising). Closing the `wait_futures` hole needs the add and the snapshot to be ordered against each other, not just made None-safe.

Not part of this finding, but adjacent and pre-existing (byte-identical in master): `Local.__getattr__/__setattr__/__delattr__` dereference `env.env_locals` with **no** None guard at all, against the unlocked null in `destroy_environment`.

src/cython/vapoursynth.pyx
```diff
@@ -2220,33 +2321,38 @@ cdef class RawNode(object):
         sidx = 0
 
         try:
-            while (not finished) or (curr_frames > 0) or running > 0:
-                if sidx not in reorder:
-                    # Spin. Reorder being empty should never happen.
+            while True:
+                with lock:
+                    if (finished and curr_frames == 0 and running == 0):
```
HolyWu: **Retracted — this comment was wrong. Please disregard it.**

Both load-bearing claims fail against `git show master:src/cython/vapoursynth.pyx`.

1. *"It also now calls `_refill()` while holding `lock`, issuing up to `prefetch` getFrameAsync calls under it."* — master already did exactly this. `master`'s `_refill` wraps the whole `while ... _request_next()` loop in `with lock:`, and `_request_next` wraps `enum_fut()` — which is what reaches `get_frame_async` → `funcs.getFrameAsync` — in `with lock:` as well. `lock` is an `RLock` in both versions, so the PR's call is a same-thread reentrant acquire and the hold time around `getFrameAsync` is unchanged.

2. *"starving the workers that are trying to deliver the very frame it waits for"* — backwards. The future is inserted into `reorder` at **request** time (`reorder[idx] = fut`, before `add_done_callback`), not at completion time. So while a frame is in flight the consumer blocks inside `fut.result()` with the lock released; the `fut is None` spin path is only reachable when `curr_frames == 0`, i.e. a momentarily empty pipeline. Master busy-waited in that same window with `if sidx not in reorder: continue`.

For the record in the other direction: the rewrite actually closes two real races in master — the `if finished: return` pre-checks in `_request_next`/`_refill` were outside the lock, and the `finally` did `finished = True` plus `reorder.copy()` completely unlocked, so a future requested concurrently could be added to `reorder` after the copy and never be closed.

Sorry for the noise.

src/cython/vapoursynth.pyx
```diff
@@ -896,30 +991,30 @@ class Error(Exception):
...
 def clear_outputs():
-    cdef dict outputs = _get_output_dict("clear_outputs")
-    outputs.clear()
+    _get_output_dict("clear_outputs").clear()
+
 
 def get_outputs():
-    cdef dict outputs = _get_output_dict("get_outputs")
-    return MappingProxyType(outputs)
+    return MappingProxyType(_get_output_dict("get_outputs").copy())
```
HolyWu: This turns `get_outputs()` from a live view into a one-shot snapshot, which contradicts the documented contract in `doc/pythonreference.rst`:

> Return a read-only mapping of all outputs registered on the current environment.
>
> **The mapping will automatically update when a new output is registered.**

A consumer that grabs `outputs = vs.get_outputs()` once and re-reads it after the script registers more outputs (vspreview does this kind of thing) now silently never sees the new nodes.

If the snapshot is intentional, the docs and ChangeLog need updating; otherwise `SynchronizedDict` should expose a real read-only live view (it already implements `__getitem__`/`__len__`/`__iter__`/`__contains__`, so `MappingProxyType` can't wrap it, but a small `Mapping` subclass delegating to it would work).

src/cython/vapoursynth.pyx
```diff
@@ -4010,26 +4140,32 @@ cdef public api int vpy4_getAvailableOutputNodes(VSScript *se, int size, int *ds
...
 
 cdef public api VSCore *vpy4_getCore(VSScript *se) nogil:
     with gil:
         try:
-            core = vsscript_get_core_internal(_get_vsscript_policy().get_environment(se.id))
-            if core is not None:
+            env = _get_vsscript_policy().get_environment(se.id)
+            if env is None:
+                return NULL
+            core = vsscript_get_core_internal(env)
```
HolyWu, line 4242: This call lands in the one lazy initialiser the PR left completely unsynchronised — `vsscript_get_core_internal`, ~line 3396:

```cython
cdef Core vsscript_get_core_internal(EnvironmentData env):
    if env.core is None:
        env.core = createCore(env)
    return env.core
```

This is the most central lazy init in the module and it sits on the very object this PR just gave an `env.lock` to, yet the lock is not taken here. Every `vs.core` access reaches it via `_get_core()`, as does `_set_logger`, `get_current_environment()`, `vpy4_getCore` and `vpy4_getVariable`.

Two threads sharing one `EnvironmentData` — e.g. a script that starts a worker thread, or a filter callback thread plus the main thread, both reaching `core.std.BlankClip` for the first time:

- T1 reads `env.core is None` → True → calls `createCore(env)` → `funcs.createCore()` builds VSCore **A** (N worker threads, framebuffer cache, full plugin scan).
- T2 reads `env.core is None` → still True → builds VSCore **B**.
- T1 stores A. T2 stores B.

Consequences, in order of severity:

1. **The orphaned core is never freed.** `destroy_environment` only frees whatever is in `env.core` at teardown, so core A leaks along with its whole thread pool and cache, for the lifetime of the process.
2. **Non-atomic refcounted store.** `env.core = createCore(env)` on a `cdef Core` attribute compiles to DECREF-old-then-store with no critical section, and `return env.core` is a separate borrowed read plus INCREF. Two threads interleaving there is exactly the "reference counting will end up inconsistent and the interpreter will crash" case from the Cython free-threading docs.
3. **Two callers get different cores.** Nodes created against A and B cannot be mixed, so the caller that lost the race gets `Error` on any cross-core operation — with no obvious cause.

Related, same function: it never consults `env.alive`, so it will happily build a brand-new `VSCore` on an environment `destroy_environment` has already torn down (`env.core = None`, `alive = False`). Anything still inside a `with env.use():` block — a logging call reaching `PythonVSScriptLoggingBridge.emit`, `_showwarning`, or a host calling `vpy4_getCore` — resurrects a core that no teardown path can ever reach again, since a second `destroy_environment` returns early on `not env.alive`.

Both want fixing together: take `env.lock` around the check-and-create (double-checked, with the second check inside the lock) and refuse to create when `not env.alive`. Note `cython.pymutex` is not reentrant, so it is worth confirming `createCore()` cannot re-enter anything that takes `env.lock` — as written it calls `funcs.createCore()` and `createCoreTimings()`, neither of which does.

src/cython/vapoursynth.pyx
```diff
@@ -4039,10 +4175,7 @@ cdef public api VSCore *vpy4_getCore(VSScript *se) nogil:
...
     return getVapourSynthAPI(version)
 
-cdef const VSAPI *getVSAPIInternal() nogil:
-    global _vsapi
-    if _vsapi == NULL:
-        _vsapi = getVapourSynthAPI(VAPOURSYNTH_API_VERSION)
+cdef inline const VSAPI *getVSAPIInternal() nogil:
     return _vsapi
 
 cdef public api int vpy4_getVariable(VSScript *se, const char *name, VSMap *dst) nogil:
```
HolyWu, line 4256: `vpy4_getVariable` is the one `vpy4_*` entry point that did not get the `if env is None` guard this PR added everywhere else. Seven lines below:

```cython
core = vsscript_get_core_internal(_get_vsscript_policy().get_environment(se.id))
```

The PR added `env = ...; if env is None: return ...` to `vpy4_getOutput`, `vpy4_getAlphaOutput`, `vpy4_getAltOutputMode`, `vpy4_getAvailableOutputNodes` and `vpy4_getCore`. Here the result of `get_environment()` goes straight into `vsscript_get_core_internal(EnvironmentData env)`, which does `if env.core is None` — a C-level attribute read on a `cdef class` with `nonecheck` off, i.e. undefined behaviour rather than a catchable exception, so the surrounding `except: return 1` does not save it.

It is a narrow race rather than an unconditional bug, because this is the *second* lookup in the function — `_vsscript_use_environment(se.id)` on the line above already raises `ValueError("env may not be None.")` if the environment is gone at that point. The window is between the two lookups:

- T1 enters `vpy4_getVariable`; the first lookup succeeds.
- T2 runs the teardown path — `_free_environment` popping the id from `_env_map`, or the atexit hook at module scope calling `clear_policy(delay=True)` → `on_policy_cleared` → `self._env_map = None`.
- T1's second lookup returns `None` → `env.core` is read off `Py_None`.

The one-line fix is to reuse the environment the function already resolved instead of looking it up twice, which closes the window and matches the shape of the other five:

```suggestion
cdef public api int vpy4_getVariable(VSScript *se, const char *name, VSMap *dst) nogil:
    with gil:
        env = _get_vsscript_policy().get_environment(se.id)
        if env is None:
            return 1
        with use_environment(env).use():
```

(Worth noting while you are in here: that `ValueError` from `_vsscript_use_environment` is raised outside the inner `try`, so it escapes a `public api ... nogil` function rather than becoming the documented `1` return. That part is pre-existing, not from this PR.)

src/cython/vapoursynth.pyx
Ichunjo: Your deduction regarding the filtering semantics in `VSScriptEnvironmentPolicy.set_environment()` is entirely accurate. Because `set_environment()` conditionally demotes any non-alive environment instance to `None` via its `is_alive()` predicate, clearing `env.alive = False` prior to invoking the `on_destroy` callbacks caused `_FastManager.__enter__` to populate the thread-local environment slot with `None`. As a consequence, any registered teardown hook touching `vs.core`, `vs.clear_outputs()`, or `vs.Local` silently threw an `Error` that was swallowed and logged to `env.core.log_message(MESSAGE_TYPE_CRITICAL)` / `sys.stderr`, precluding proper cleanup.

Furthermore, as you noted, prematurely zeroing `env.alive` exposed an external race where a secondary concurrent `destroy_environment()` invocation would observe `not env.alive` and return instantaneously—signaling completed teardown to its caller while the primary thread was still midway through `wait_futures()`, user callbacks, and C API `freeCore()` invocations, thereby creating a catastrophic double-free window in host embeddings.

To resolve both hazards without sacrificing atomic idempotency:
1. We introduced a dedicated `cdef bint destroying` state flag on `EnvironmentData`.
2. Upon entering `destroy_environment()`, the critical section under `env.lock` atomically validates `if not env.alive or env.destroying: return` and transitions `env.destroying = True`.
3. `env.alive` remains `True` throughout the entire callback dispatch phase, permitting `VSScriptEnvironmentPolicy` to properly bind the environment on the executing thread.
4. Only after all callbacks finish, node/frame invalidation sweeps complete, and `freeCore()` finishes is `env.alive` transitioned to `False` under `env.lock` alongside container neutralization.

src/cython/vapoursynth.pyx
Ichunjo: Spot-on analysis. Under default Cython compilation semantics with `nonecheck(False)`, accessing C-level extension attributes on a `cdef class SynchronizedDict`—specifically `self.active_exceptions.lock` and `self.active_exceptions.data`—compiles directly into raw struct offset pointer arithmetic without emitting a `Py_None` type check. 

Because `destroy_environment()` concurrently cleared `self.active_exceptions = None` without synchronization against `store_exception()`, the unlocked initial check `if self.active_exceptions is None:` was vulnerable to a time-of-check to time-of-use (TOCTOU) race. If the field was nulled in that window, Cython dereferenced `((SynchronizedDict *)Py_None)->lock`, executing a `PyMutex_Lock` over arbitrary memory within or adjacent to the `Py_None` singleton, precipitating an uncatchable segmentation fault.

We have remediated this by acquiring a local strong reference prior to inspection:
```cython
cdef SynchronizedDict exc_dict = self.active_exceptions
if exc_dict is None:
    return 0
with exc_dict.lock:
    ...
```
This guarantees an atomic `Py_INCREF` on the underlying `SynchronizedDict` instance; even if concurrent teardown drops `self.active_exceptions`, the local reference keeps the instance and its embedded `cython.pymutex` alive and valid throughout the lock acquisition.

src/cython/vapoursynth.pyx
Ichunjo: Agreed. Enforcing mutual exclusion exclusively on the consumer/producer read sites (`createVideoNode`, `createVideoFrame`, `createAudioNode`, etc.) while leaving the teardown writer block completely unguarded established an asymmetric synchronization discipline that offered zero race protection under free-threading.
Specifically, as you delineated:
1. A thread acquiring `env.lock` inside `createVideoNode` could pass `if env.known_nodes is not None:`, only for `destroy_environment` to execute `env.known_nodes = None` concurrently without holding `env.lock`, causing an unhandled `AttributeError` inside `.add()`.
2. A thread constructing a node immediately subsequent to the invalidation sweep would insert a new `VSNode*` into `known_nodes` just as `freeCore()` severed the underlying core, producing an immediate use-after-free upon the subsequent frame request.
We have unified the locking discipline: `destroy_environment` now acquires `with env.lock:` both when snapshotting the `known_nodes`/`known_frames` collections for teardown and when ultimately resetting `outputs`, `core`, `env_locals`, `active_exceptions`, `known_nodes`, and `known_frames` to `None`.

src/cython/vapoursynth.pyx
Ichunjo: Your updated assessment precisely identifies the failure mode of the classic double-checked locking anti-pattern in `get_policy()`. 

Because `EnvironmentPolicyAPI.ensure_policy_matches()` strictly validates identity via `_policy is self._target_policy()`, `_policy = policy` must necessarily precede `policy.on_policy_registered(_api)` (which invokes `api.create_environment()`). Consequently, the fast path in `get_policy()`—which inspected `if _policy is None:` without synchronization—allowed concurrent threads to observe a non-`None` `_policy` while `StandaloneEnvironmentPolicy.on_policy_registered()` was still midway through `set_logger()` and initializing `self._environment` (including the full dynamic plugin scan). Any thread hitting that multi-millisecond window returned an uninitialized policy whose `get_current_environment()` evaluated to `None`.

Furthermore, as you astutely noted, making `_policy_lock` unconditional inside `get_policy()` was untenable due to the non-reentrant nature of `cython.pymutex`.
We have resolved this via reader-side readiness gating using a synchronization event:
1. Introduced a module-level `_policy_ready = threading.Event()`.
2. In `_register_policy_unlocked()`, `_policy_ready.clear()` is asserted before assigning `_policy = policy`. Upon successful return of `policy.on_policy_registered(_api)` (or within a `finally` block on failure), `_policy_ready.set()` is signaled.
3. In `get_policy()`, the fast path checks `if not _policy_ready.is_set():` before falling back to acquiring `_policy_lock` and invoking `_policy_ready.wait()`.
4. `clear_policy()` explicitly clears `_policy_ready`.
This closes the publication-before-initialization race while completely avoiding non-reentrant mutex re-acquisition deadlocks.

pyproject.toml
Ichunjo: Acknowledged and appreciated. We verified against the `cibuildwheel` 4.2.0 codebase that `cpython-freethreading` is not an `EnableGroup` enum member, and that `cp315t-*` target identifiers are active by default without auxiliary gating configuration. The `pyproject.toml` specification has been preserved as written.

src/cython/vapoursynth.pyx
Ichunjo: Thank you for the clarification regarding `WeakSet` internals on free-threaded 3.13+ (where per-object locking on the underlying set provides atomic `add` operations).

The fundamental vulnerability you highlighted—the ordering divergence between `active_futures.add(fut)` and the teardown snapshot `futs = env.active_futures`—was indeed critical. If a producer thread in `VideoNode.get_frame`, `AudioNode.get_frame`, or `CallbackData.__init__` evaluated `is not None` and executed `.add()` *after* `destroy_environment()` had already snapshotted `futs` and set `env.active_futures = None`, two catastrophic outcomes emerged:
1. An `AttributeError` escaped the request call if `env.active_futures` became `None` mid-expression.
2. If the future was appended to an orphaned set after the snapshot, `wait_futures(futs)` returned prematurely. Teardown then invoked `funcs.freeNode(rn.node)` across all active nodes while the worker thread was actively executing `with nogil: funcs.getFrame(...)`, culminating in a use-after-free on the C `VSNode*`.

To guarantee absolute serializability, we wrapped all producer additions (`CallbackData.__init__`, `VideoNode.get_frame`, `AudioNode.get_frame`) in `with env.lock: if env.active_futures is not None: env.active_futures.add(fut)`. Because `destroy_environment()` snapshots and clears `env.active_futures` under that identical `env.lock`, a frame request is either guaranteed to be included in the `wait_futures()` barrier or is rejected because teardown has already commenced.

src/cython/vapoursynth.pyx
Ichunjo: Acknowledged regarding the comparison with `master`. 

While the locking hold time in `_refill()` was pre-existing, we took this opportunity to replace the underlying busy-spin polling loop (`if sidx not in reorder: continue`) with a standard `threading.Condition(Lock())` synchronization mechanism (`cond.wait()` / `cond.notify_all()`). Under free-threaded execution where GIL time-slicing cannot be relied upon to yield execution, the `Condition` variable prevents the consumer thread from burning 100% CPU on an idle core while waiting for out-of-order frame delivery from upstream filter workers.

src/cython/vapoursynth.pyx
Ichunjo: You are entirely correct. Wrapping a snapshot copy via `MappingProxyType(outputs.copy())` broke the long-standing contract documented in `doc/pythonreference.rst` stating that the returned mapping dynamically reflects newly registered or modified outputs over the lifetime of the environment.

Because `MappingProxyType` cannot wrap arbitrary Cython extension types lacking a standard C-level dict mapping proxy interface, we implemented a dedicated `SynchronizedDictView` implementing `collections.abc.Mapping`:
```cython
class SynchronizedDictView(Mapping):
    def __init__(self, target):
        self._target = target

    def __getitem__(self, key):
        return self._target[key]

    def __iter__(self):
        return iter(self._target)

    def __len__(self):
        return len(self._target)

    def __repr__(self):
        return f"{type(self).__name__}({dict(self._target.items())!r})"
```

`get_outputs()` now returns `SynchronizedDictView(_get_output_dict("get_outputs"))`, fully restoring live view semantics while delegating every access to thread-safe `SynchronizedDict` methods under **`PyMutex.`**

src/cython/vapoursynth.pyx
Ichunjo: This is an exceptional finding. `vsscript_get_core_internal()` was indeed the single most vulnerable unsynchronized lazy initializer in the codebase. 

Under free-threading without the GIL, two threads concurrently resolving `vs.core` (or calling `_set_logger` / `vpy4_getCore`) on an uninitialized environment observed `env.core is None` simultaneously. Both proceeded to invoke `createCore(env)`, generating two independent `VSCore` instances complete with separate thread pools, memory caches, and dynamic plugin enumerations. Because `env.core = createCore(env)` was assigned non-atomically, one core was irreversibly orphaned and leaked for the lifetime of the process, and the non-atomic store/decref on `cdef Core` invited reference counting corruption. 

Worse, because `vsscript_get_core_internal` omitted any lifecycle verification, any lingering reference (such as a logging bridge emit during process termination or a host querying `vpy4_getCore`) resurrected a brand-new `VSCore` on an already-destroyed environment (`env.alive == False`), which could never subsequently be collected because subsequent `destroy_environment()` calls returned early on `not env.alive`.

We have remediated this by implementing proper double-checked locking guarded by `env.lock` alongside explicit lifecycle validation:
```cython
cdef Core vsscript_get_core_internal(EnvironmentData env):
    cdef Core core = env.core
    if core is None:
        with env.lock:
            if not env.alive or env.destroying:
                raise Error('Cannot access or create core on a destroyed environment.')
            if env.core is None:
                env.core = createCore(env)
            core = env.core
    return core
```

Because `createCore()` exclusively executes `funcs.createCore()` and initializes core timing structures without re-entering `EnvironmentData` lock primitives, `cython.pymutex` recursion is completely avoided.

src/cython/vapoursynth.pyx
Ichunjo: Completely agreed. `vpy4_getVariable` suffered from both an unprotected lookup sequence and a redundant secondary query.

Specifically:
1. `with _vsscript_use_environment(se.id).use():` was placed outside the internal `try...except` block; if the environment was missing or already destroyed, `_vsscript_use_environment` raised a Python `ValueError("env may not be None.")` that escaped a `public api ... nogil` C export, violating the function's documented C return error protocol.
2. Evaluating `core = vsscript_get_core_internal(_get_vsscript_policy().get_environment(se.id))` constituted a second lookup against `_env_map`. If concurrent teardown or an atexit hook cleared the environment between lookup 1 and lookup 2, `get_environment()` returned `None`, passing `Py_None` directly into `vsscript_get_core_internal(EnvironmentData env)` and triggering an unchecked struct dereference.

We have refactored both `vpy4_getVariable` and `vpy4_setVariables` to resolve the environment exactly once, guard against `None` inside the `try...except` boundary, and reuse the resolved `env` instance:
```cython
cdef public api int vpy4_getVariable(VSScript *se, const char *name, VSMap *dst) nogil:
    with gil:
        try:
            env = _get_vsscript_policy().get_environment(se.id)
            if env is None:
                return 1
            with use_environment(env).use():
                pyenvdict = <dict>se.pyenvdict
                dname = name.decode('utf-8')
                read_var = { dname: pyenvdict[dname] }
                core = vsscript_get_core_internal(env)
                if core is None or (<Core>core).core == NULL:
                    return 1
                dictToMap(read_var, dst, (<Core>core).core, (<Core>core).funcs)
                return 0
        except:
            return 1
```

In tandem, we have deleted the redundant and now-dead `_vsscript_use_environment` and `_vsscript_use_or_create_environment` helper functions entirely.

myrsloik:
Could the just "improving things" bits be split out so I can merge them into the vukvuk2 branch already? Things like the collect_garbage paramter, most CI changes and decorators.

Ichunjo:
Created PRs https://github.com/vapoursynth/vapoursynth/pull/1259 https://github.com/vapoursynth/vapoursynth/pull/1260 and https://github.com/vapoursynth/vapoursynth/pull/1261

PR #1262, opened 2026-08-28
Add support for Python free-threading, second iteration

Ichunjo:
Second iteration of #1257

- Configured Meson and headers to support free-threaded builds (`Py_GIL_DISABLED`) without limited API (abi3t isn't supported yet), while keeping abi3 for standard Python. Bumped Cython requirement to >= 3.3.0.
- Added a `SynchronizedDict` helper (`cython.pymutex`) and locked access to environment maps, known nodes, and async frame reordering queues.
- Initialized `_vsapi` at import time to prevent lazy-init race conditions.
- Replaced manual counter / `WeakValueDictionary` with `weakref.WeakSet`.

### DIFF
--- a/include/cython/vapoursynth_api.h
+++ b/include/cython/vapoursynth_api.h
@@ -7,7 +7,11 @@
 #define MS_WIN64
 #endif
 
+#include "pyconfig.h"
+
+#ifndef Py_GIL_DISABLED
 #define Py_LIMITED_API 0x030C0000
+#endif
 
 #include "Python.h"
 #include "vapoursynth.h"
--- a/meson.build
+++ b/meson.build
@@ -300,12 +300,16 @@ vapoursynth_dep = declare_dependency(
     link_with: libvapoursynth,
 )
 
+is_freethreaded = py.get_variable('Py_GIL_DISABLED', 0) == 1 or py.get_variable('ABIFLAGS', '').contains('t')
+py_limited_api = is_freethreaded ? '' : '3.12'
+
 py.extension_module('vapoursynth',
     files('src/common/wave.cpp', 'src/cython/vapoursynth.pyx'),
+    cython_args: ['-X', 'freethreading_compatible=True'],
     dependencies: [py_dep, vapoursynth_dep, vulkan_headers],
     install: true,
     install_rpath: '$ORIGIN',
-    limited_api: '3.12',
+    limited_api: py_limited_api,
     override_options: 'cython_language=cpp',
     subdir: 'vapoursynth',
 )
@@ -324,7 +328,7 @@ py.install_sources(
 
 libvsscript = library('vsscript',
     files('src/vsscript/vsscript.cpp'),
-    cpp_args: ['-DPy_NO_LINK_LIB', '-DVS_CORE_EXPORTS'],
+    cpp_args: ['-DPy_NO_LINK_LIB', '-DVS_CORE_EXPORTS'] + (is_freethreaded ? ['-DPy_GIL_DISABLED=1'] : []),
     dependencies: [dl_dep, py_dep, threads_dep],
     gnu_symbol_visibility: 'hidden',
     include_directories: incdir,
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["Cython>=3.1.0", "meson-python"]
+requires = ["Cython>=3.3.0", "meson-python"]
 build-backend = "mesonpy"
 
 [project]
@@ -9,9 +9,7 @@ readme = "README.md"
 requires-python = ">=3.12"
 license = "LGPL-2.1-or-later"
 license-files = ["COPYING.LESSER"]
-authors = [
-    {name = "Fredrik Mellbin", email = "fredrik.mellbin@gmail.com"},
-]
+authors = [{ name = "Fredrik Mellbin", email = "fredrik.mellbin@gmail.com" }]
 keywords = ["frameserver", "video", "audio"]
 classifiers = [
     "Development Status :: 5 - Production/Stable",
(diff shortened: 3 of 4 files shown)


myrsloik:
image

Ichunjo:
Adressed the two bugs.

3. ?
4. I can add cp314t if you really want it but I don't see why you would want it when 3.15 is about to be released.
5. Fixed the unused Mapping/sentinel is a thing and does exist.
6. Yes

myrsloik:
One more round if AI torture for you:

One new finding, introduced by the fix to #1 — a concurrent waiter can now hang forever:

Thread A claims a user policy (_policy = P, event cleared). Thread B enters get_policy(): event unset, not the registering thread, and — the key step — _policy is not None, so B skips the create-standalone branch and blocks in _policy_ready.wait(). A's on_policy_registered raises → the rollback sets _policy = None and clears the already-clear event. Nothing ever wakes B.

This isn't fixable by flipping the failure path to set() — that just poisons instead: B would wake into if _policy is None: raise RuntimeError, and every later get_policy() would skip the whole block (event set) and hit the same raise, with the auto-standalone fallback permanently dead. The root cause is that an Event can't express a state that moves backwards. The right shape is a Condition with a predicate loop — wait for "ready or _policy is None", have the failure path notify_all, and let woken waiters loop back to the create-standalone branch. (The condition's lock can subsume _policy_lock while it's at it.)

A smaller gap in the same function: the EnvironmentPolicyAPI construction — including weakref.ref(_policy) — sits between the claim and the try. A non-weakrefable policy (__slots__ without __weakref__ is enough) raises TypeError there, leaving _policy claimed, the event cleared, and _registering_thread stuck set (the finally doesn't cover that region) — after which every other thread's get_policy() hangs. The old code failed here too (differently), but the rollback machinery now exists and should simply cover everything after the claim: move the API construction inside the try.



Nit: cdef bool _register_policy(...) — bare bool in Cython is the Python type (there's no from libcpp cimport bool here), so this compiles but boxes; bint is the idiom.

Ichunjo:
Fixed

espressolee (changes requested):
Reviewed exact head `c0dce6cdd6d8c76925a896af0025403a07278d9d` using its macOS arm64 CI artifact `9716817596`; the downloaded archive digest matches GitHub's `sha256:9810cf148d9d7865b9bb6236cd35e17715a8d550383c5a8d4ad4975a7312df3a`. The CPython 3.15.0b4t extension imports with `PYTHON_GIL=0` and leaves the GIL disabled.

I found two lifecycle blockers and one public mapping-semantics regression:

1. **[P1] Teardown can still win before an operation is admitted, after which the operation uses the invalidated native pointer** (`src/cython/vapoursynth.pyx:1154-1159`, `2601-2612`, and the analogous audio path at `2939-2950`). The new code locks the `active_futures` add, but `None` is treated as "skip tracking and continue" rather than "teardown has claimed this environment; abort". I made the window deterministic by replacing the module-level `Future` with a subclass that pauses immediately after `set_running_or_notify_cancel()`: one thread passes `ensure_valid_frame_number()` and pauses, another calls `destroy_environment()`, then the first resumes. At this head under `PYTHON_GIL=0`, all 5/5 subprocess trials terminated with SIGSEGV in `VideoNode.get_frame()` because destroy had already NULLed/freed the node/core and the reader still called native `getFrame()`.

   Exact base also has this race, so I am not calling it a regression introduced here. It is nevertheless still open in the exact lifecycle surface this PR is synchronizing. Admission needs to be atomic with `alive/destroying` under `env.lock`: if teardown wins, reject before native use; if the operation wins, teardown must see and wait for it. For synchronous `get_frame()`, the guard also needs to remain live through error decoding and `createConst*Frame()`/frame registration, not be completed immediately after the native `getFrame()` call. `CallbackData` needs the same fail-closed admission before `getFrameAsync()` is submitted.

2. **[P1] A policy can unregister itself during `on_policy_registered()` and leave `_policy_ready=True` with `_policy=None`** (`src/cython/vapoursynth.pyx:657-661`). `EnvironmentPolicyAPI.unregister_policy()` is available to the registration callback. If the callback invokes it, `clear_policy()` clears `_policy`; when the callback returns, `_register_policy()` unconditionally publishes readiness. At exact base, the same policy leaves no registered policy and a later `get_current_environment()` correctly creates the standalone policy. At this head, in both GIL modes, `has_policy()` is false but `get_current_environment()` fails with `AttributeError: 'NoneType' object has no attribute 'get_current_environment'` because the fast path sees `_policy_ready`. The success transition needs to verify the same policy/registration generation is still claimed, or unregister-during-registration must be rejected and rolled back as one state-machine transition.

3. **[P2] `get_outputs()` no longer preserves mapping-view behavior** (`src/cython/vapoursynth.pyx:290-300`). Exact base returns `dict_keys`, `dict_items`, and `dict_values` from its documented auto-updating read-only mapping. At this head, those methods return plain lists. In a same-Python 3.14.6 base/head comparison, a saved `keys = get_outputs().keys()` saw a later `set_output(37)` on base, but not on this head; saved items behaved the same way. Please keep the synchronized snapshotting without changing the public view contract (or explicitly change/document the API if that is intended).

Existing focused tests remain green here: `environment_test.py` 12/12 under both GIL modes and `async_test.py` 4/4 under `PYTHON_GIL=0`. The exact-head Actions jobs are successful builds/codespell, but the workflow does not execute these runtime tests, so those green checks do not exercise the failures above.

Ichunjo:
I will no longer accept sloppy reviews from clankers. If I want to have a chat with one, I can do that myself

myrsloik:
Maybe environment policies need to be slightly modified somehow to sidestep/mitigate some issues. Or maybe threading simply is hard. It always is.

myrsloik:
Good enough after real review

PR #1263, opened 2026-08-28
Prevent frame leaks in RawNode.frames()/RawNode.output 

Ichunjo:
- Refactor the frame enumeration logic in `RawNode` to avoid using a generator within a generator, which was causing issues with frame lifecycle management when an exception was thrown from a ModifyFrame callback or from the progress_update callback.
- Add explicit `try...finally` blocks in `VideoNode` and `AudioNode` to guarantee that the frame generator is closed and buffers are freed, even when exceptions occur during iteration.

`_get_enum_fut` caused a memory leak because it creates a secondary generator object.
When `frames()` is cancelled or interrupted early, it remains in a suspended state holding references to `RawNode` and `get_frame_async`.
Apparently, in Cython, nesting a generator inside another generator generates cross-referencing scope structs that create reference cycles that prevent GC deallocation.

The entire core and everything else, was being left orphaned in memory with no way to free it.

---

Draft for now, until #1262 is merged

### DIFF
--- a/src/cython/vapoursynth.pyx
+++ b/src/cython/vapoursynth.pyx
@@ -2391,18 +2391,9 @@ cdef class RawNode(object):
             backlog = prefetch * 3
         backlog = max(backlog, prefetch)
 
-        def _get_enum_fut():
-            get_frame_async = self.get_frame_async
-
-            for n in range(self.num_frames):
-                handle = HandleFuture()
-                get_frame_async(n, handle)
-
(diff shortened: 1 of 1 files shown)


myrsloik:
Rebase it and fix this and I'll merge it:
5	Nit	Mapping is added to the collections.abc import but never used anywhere in the diff.

Ichunjo:
Rebased. Should be ready to be merged now.

PR #1266, opened 2026-09-20
Fix mkv muxing bug in vspipe

no1d:
When MuxStream is pushed into vector the pointer in the stored request can become invalid.
The easy fix would be storing pointers to streams, but having a pointer to self seems quite error prone to me.


myrsloik:
A slightly different fix applied

